PR #54, opened 2015-01-23
export classy chalk object to create isolated ctx

### LINKED ISSUE #46: `enabled` is global state
Using chalk from a library I know I always want colours enabled as it is explicitly requested to begin with. But doing `require('chalk').enabled = true` would be bad pratice as this enforce the setting globally and possibly interfere with other libraries using chalk and depending on the auto-detection.

It would be useful to be able to set this option in some isolated way.

keis:
First stab at implementing something like discussed in #46 

wdyt?

-

fixes #46

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,13 @@ var stripAnsi = require('strip-ansi');
 var hasAnsi = require('has-ansi');
 var supportsColor = require('supports-color');
 var defineProps = Object.defineProperties;
-var chalk = module.exports;
+
+function Chalk(enabled) {
+	// detect mode if not set manually
+	if (enabled === undefined) {
+		this.enabled = this.supportsColor;
+	}
+}
 
 // use bright blue on Windows as the normal blue color is illegible
 if (process.platform === 'win32') {
@@ -17,6 +23,7 @@ function build(_styles) {
 		return applyStyle.apply(builder, arguments);
 	};
 	builder._styles = _styles;
+	builder.enabled = this.enabled;
 	// __proto__ is used because we must return a function, but there is
 	// no way to create a function with a different prototype.
 	builder.__proto__ = proto;
@@ -31,7 +38,7 @@ var styles = (function () {
 
 		ret[key] = {
 			get: function () {
-				return build(this._styles.concat(key));
+				return build.call(this, this._styles.concat(key));
 			}
 		};
 	});
@@ -53,7 +60,7 @@ function applyStyle() {
 		}
 	}
 
-	if (!chalk.enabled || !str) {
+	if (!this.enabled || !str) {
 		return str;
 	}
 
@@ -78,7 +85,7 @@ function init() {
 	Object.keys(styles).forEach(function (name) {
 		ret[name] = {
 			get: function () {
-				return build([name]);
+				return build.call(this, [name]);
 			}
 		};
 	});
@@ -86,14 +93,11 @@ function init() {
 	return ret;
 }
 
-defineProps(chalk, init());
+defineProps(Chalk.prototype, init());
 
-chalk.styles = ansiStyles;
-chalk.hasColor = hasAnsi;
-chalk.stripColor = stripAnsi;
-chalk.supportsColor = supportsColor;
+Chalk.prototype.styles = ansiStyles;
+Chalk.prototype.hasColor = hasAnsi;
+Chalk.prototype.stripColor = stripAnsi;
+Chalk.prototype.supportsColor = supportsColor;
 
-// detect mode if not set manually
-if (chalk.enabled === undefined) {
-	chalk.enabled = chalk.supportsColor;
-}
+module.exports = new Chalk();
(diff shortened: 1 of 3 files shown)


sindresorhus:
@keis Thanks for doing this. I've been neglecting making a decision on this for so long.

Would you mind running the benchmark (`npm run bench`) before and after your changes and paste it here?

// @jbnicolai @julien-f @UltCombo @stevenvachon @SBoudrias @naartjie

keis:
Sure thing.

@ master
> chalk@0.5.1 bench /home/keis/git/chalk
> matcha benchmark.js


                      chalk
       1,565,140 op/s » single style
         328,343 op/s » several styles
       1,941,015 op/s » cached styles
         349,328 op/s » nested styles


  Suites:  1
  Benches: 4
  Elapsed: 5,833.32 ms

@ isolated
> chalk@0.5.1 bench /home/keis/git/chalk
> matcha benchmark.js


                      chalk
       1,459,222 op/s » single style
         321,070 op/s » several styles
       2,014,983 op/s » cached styles
         335,849 op/s » nested styles


  Suites:  1
  Benches: 4
  Elapsed: 6,085.44 ms

UltCombo:
Looks nice!
Perhaps export the constructor so one can do `new chalk.Chalk()`? It'd look a bit nicer than accessing the default export's `constructor` property IMHO.

index.js
```diff
@@ -86,14 +93,11 @@ function init() {
...
 }
 
-defineProps(chalk, init());
+defineProps(Chalk.prototype, init());
 
-chalk.styles = ansiStyles;
-chalk.hasColor = hasAnsi;
-chalk.stripColor = stripAnsi;
-chalk.supportsColor = supportsColor;
+Chalk.prototype.styles = ansiStyles;
```
SBoudrias: I don't think any of these needs to be on the prototype, make them Class props.

keis: The idea is to get the names exported so I guess I could just stick them on exported object. Would then change the reference to `this.supportsColor` to just use the local supportsColor as I'm guessing there is not really much use in being able to override that.

UltCombo:
On second thought, seeing as the default export is an instance (which makes sense for back-compat), `new chalk.constructor()` may feel more natural than exposing the constructor as (another) property of the default instance (as I had suggested `new chalk.Chalk()` previously).

Though, the fact that the default export is an instance is pretty much just an implementation detail, in that case `new chalk.Chalk()` would make more sense.

I guess it doesn't matter much either way.

sindresorhus:
@keis This looks good to me. Would you mind adding docs?

Something like:

> The `enabled` option should only be used by the end-user as it affects all chalk consumers. Create a new instance if you need to change this just for your usage.

keis:
@UltCombo I think using the constructor is kinda elegant. But I might be a bit blinded by the novelty of it

@sindresorhus would be good to add this to the chalk.enabled section I guess

>### chalk.enabled
>Color support is automatically detected, but you can override it by setting the
>`enabled` property or by creating a new instance just for your usage. Changing
>the property should only be done from end-user facing applications as it
>affects all chalk consumers.

UltCombo:
I guess how the constructor is exported doesn't matter much.

The text looks a bit confusing though. When you put the "Changing the property should only be done from end-user" as the last sentence it looks like you're talking about all instances. I guess that part should be explicit for the default instance, plugins that create their own instances can do as they please.

keis:
hmm, good point.

"Changing the property should only be done from end-user facing applications as it
affects all consumers of the default chalk instance" maybe?

UltCombo:
Sounds good =]

index.js
sindresorhus: :+1:

test.js
```diff
@@ -68,6 +68,14 @@ describe('chalk.enabled', function () {
 	});
 });
 
+describe('chalk.constructor', function () {
+	it('should create a isolated context where colors can be disabled', function () {
+		var ctx = new chalk.constructor(false);
```
sindresorhus: I don't think passing `false` into the constructor is very intuitive. Wouldn't it be better to just create a new instance and then change the `enabled` property?

```js
var ctx = new chalk.constructor();
ctx.enabled = false;
```

keis: Agreed, that looks pretty ugly. But I do like being able to create the instance fully configured in one go.

`new chalk.constructor({enabled: false})` is perhaps cleaner

sindresorhus: :+1:

index.js
sindresorhus: @keis Would you mind resolving this so the PR can be landed? :)

keis: yep, I'll get that fixed. thanks poking me was getting was almost forgetting about this :)

keis:
My turn to bump this :)

just added the properties to exports and just now fixed a silly whitespace error

jbnicolai:
Big thumbs up from me :+1:

@sindresorhus any thoughts?

sindresorhus:
Landed! Thanks for pulling through this @keis. We really appreciate it :D

superhighfive

keis:
high five

PR #60, opened 2015-04-17
Don't override windows blue when in xterm mode

### LINKED ISSUE #59: (Windows) Consider using standard color scheme when term is xterm like
perhaps check for:

`process.platform === 'win32' && !process.env.TERM`
or
`process.platform === 'win32' && !(/xterm/).test(process.env.TERM || "")`

I am suggesting this because I use node through a solarized color shell (admittedly through cygwin ssh).

Thanks.

weikinhuang:
fixes #59

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ function Chalk(options) {
 }
 
 // use bright blue on Windows as the normal blue color is illegible
-if (process.platform === 'win32') {
+if (process.platform === 'win32' && !/^xterm/i.test(process.env.TERM)) {
 	ansiStyles.blue.open = '\u001b[94m';
 }
 
--- a/test.js
+++ b/test.js
@@ -1,5 +1,8 @@
 'use strict';
 var assert = require('assert');
+var requireUncached = require('require-uncached');
+var resolveFrom = require('resolve-from');
+var semver = require('semver');
 var chalk = require('./');
 
 describe('chalk', function () {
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -12,7 +13,7 @@ function Chalk(options) {
 }
 
 // use bright blue on Windows as the normal blue color is illegible
-if (process.platform === 'win32') {
+if (process.platform === 'win32' && !xtermRegexp.test(process.env.TERM || 'dumb')) {
```
sindresorhus:
```js
if (process.platform === 'win32' && !/^xterm/i.test(process.env.TERM)) {
```

jbnicolai: Fully agree that there's no need to declare the regex. This is not a hot code path (very much not so)

weikinhuang: ok, i'll make that change then

jbnicolai:
@weikinhuang appreciate the effort, and happy to merge, but would you be willing to add a (small) test asserting the correct behaviour?

weikinhuang:
@jbnicolai but the test will only work in windows

index.js
```diff
@@ -12,7 +12,7 @@ function Chalk(options) {
 }
 
 // use bright blue on Windows as the normal blue color is illegible
-if (process.platform === 'win32') {
+if (process.platform === 'win32' && !(/^xterm/i).test(process.env.TERM || 'dumb')) {
```
sindresorhus: Please copy paste what I commented, the regex doesn't need parens and `dumb` check is invalid.

jbnicolai:
@weikinhuang `process.env` can be mocked. See https://github.com/sindresorhus/supports-color/blob/master/test.js for an example of how to test emulating different environments.

jbnicolai:
@weikinhuang and last but not least: it'd be nice if you'd squash the pr into a single commit, although I'd be happy to do this for you.

weikinhuang:
@jbnicolai Will do, I like keeping separate commits right up until the point of completion, then squashing afterwards usually. Thanks for the comments!

index.js
weikinhuang: will do.

weikinhuang:
@jbnicolai Actually I am unable to mock `process.env` because that check is not a run time override. Would any of you be able to provide some tips for mocking that?

weikinhuang:
Nevermind, figured it out, had to use `require-uncached`.

jbnicolai:
Precisely :)

weikinhuang:
Updated + tests

weikinhuang:
However I think this would fail since I can't mock `process.platform` because it's a read only property. Should I just run the test when the platform is `win32`?

sindresorhus:
You can mock it with `Object.defineProperty(process, 'platform', {value: 'win32'});`.

weikinhuang:
Cool, I always forget about `Object.defineProperty`

weikinhuang:
@sindresorhus do you know why the test is failing on `v0.10`?

sindresorhus:
I think it has something to do with us overriding `process.platform` and module path resolution logic on Node 0.10. I think we can just `if` check if it's node 0.12 or newer for those tests.

weikinhuang:
Updated

test.js
```diff
@@ -1,6 +1,9 @@
 'use strict';
 var assert = require('assert');
 var chalk = require('./');
+var requireUncached = require('require-uncached');
+var resolveFrom = require('resolve-from');
+var semver = require('semver');
```
kevva: Can you move these up before `var chalk = require('./');` and after `var assert = require('assert');`?

weikinhuang: sure

weikinhuang:
Updated

sindresorhus:
Looks good. Thanks :)

sindresorhus:
Will do a new release when https://github.com/sindresorhus/chalk/issues/58 is resolved. In the meantime you can use it with `npm install --save sindresorhus/chalk` ;)

PR #67, opened 2015-05-24
.chalkrc

renegat4:
I use Solarized terminal colours. Chalk.grey() is invisible.

So i added code to have a .chalkrc file where i can overwrite colours with other colours:

```json
{
  "overwrite": {
    "grey": "blue",
    "gray": "blue"
  }
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,15 +10,15 @@ var fs = require('fs');
 function Chalk(options) {
 	// detect mode if not set manually
 	this.enabled = !options || options.enabled === undefined ? supportsColor : options.enabled;
-  var configPath = process.cwd() + '/.chalkrc';
-  if (fs.existsSync(configPath)) {
-    var content = JSON.parse(fs.readFileSync(configPath));
-    if (content.overwrite) {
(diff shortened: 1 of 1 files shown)


jbnicolai:
This seems really obscure. If your colorscheme doesn't allow you to see a color, change the scheme - not the application outputting the color.

renegat4:
Changing the colourtheme is not an option. Changing the colour breaks the theme.

The correct way to solve the problem would be to rewrite jscs to not use hardcoded chalk.grey() but rather something like 'colour1()' and make it configurable.
My approach seems far less complex. I know its a hack.

arthurvr:
But at the end what you're describing is a bug in the theme, not of chalk and not of any CLI that uses chalk.

sindresorhus:
@renegat4 This is a well known problem with Solarized. Unfortunately Solarized is not maintained anymore, so it's still an issue. There are Solarized forks that fixes this. Please don't bother tools about this. It's a problem with your theme, not the tools you're using.

PR #72, opened 2015-07-01
Makes dim when gray a noop on windows terminals.

### LINKED ISSUE #58: (Windows) chalk.gray.dim not visible
Only when highlight in cmder.

In cmd:

|Normal|Highlight|
|---|---|
|image|image|

In cmder:

|Normal|Highlight|
|---|---|
|image|image|

jbnicolai:
Fixes #58 

Could use a quick sanity check.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ var stripAnsi = require('strip-ansi');
 var hasAnsi = require('has-ansi');
 var supportsColor = require('supports-color');
 var defineProps = Object.defineProperties;
+var isSimpleWindowsTerm = process.platform === 'win32' && !/^xterm/i.test(process.env.TERM);
 
 function Chalk(options) {
 	// detect mode if not set manually
@@ -12,7 +13,7 @@ function Chalk(options) {
 }
 
 // use bright blue on Windows as the normal blue color is illegible
-if (process.platform === 'win32' && !/^xterm/i.test(process.env.TERM)) {
+if (isSimpleWindowsTerm) {
 	ansiStyles.blue.open = '\u001b[94m';
 }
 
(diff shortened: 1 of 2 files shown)


Qix-:
lgtm. Shame windows has mediocre support.

index.js
```diff
@@ -69,6 +69,15 @@ function applyStyle() {
 	var nestedStyles = this._styles;
 	var i = nestedStyles.length;
 
+	// Turns out that on Windows dimmed gray text becomes invisible in cmd.exe,
+	// see https://github.com/chalk/chalk/issues/58
+	// If we're on Windows and we're dealing with a gray color, temporarily make 'dim' a noop.
+	var originalDim = ansiStyles.dim.open;
+	if ((nestedStyles.indexOf('gray') != -1 || nestedStyles.indexOf('grey') != -1) &&
+			process.platform === 'win32' && !/^xterm/i.test(process.env.TERM)) {
```
sindresorhus: Dry this up into a top level variable.

This check should also be first, so the engine can skip the other tests.

index.js
```diff
@@ -69,6 +69,15 @@ function applyStyle() {
 	var nestedStyles = this._styles;
 	var i = nestedStyles.length;
 
+	// Turns out that on Windows dimmed gray text becomes invisible in cmd.exe,
+	// see https://github.com/chalk/chalk/issues/58
+	// If we're on Windows and we're dealing with a gray color, temporarily make 'dim' a noop.
+	var originalDim = ansiStyles.dim.open;
+	if ((nestedStyles.indexOf('gray') != -1 || nestedStyles.indexOf('grey') != -1) &&
```
sindresorhus: `!=` => `!==`

index.js
```diff
@@ -69,6 +69,15 @@ function applyStyle() {
...
+			process.platform === 'win32' && !/^xterm/i.test(process.env.TERM)) {
+		ansiStyles.dim.open = '';
```
sindresorhus: Should we reset the `.close` too so not to introduce a moot dim reset escape?

jbnicolai: Considered that. Don't really think it matters, decided to go for the simplest solution.

index.js
jbnicolai: :+1:

index.js
jbnicolai: Check.

PR #82, opened 2015-09-12
supports-color@3.1.1

hughsk:
Currently chalk breaks when run in Electron's renderer. Upgrading to `supports-color@3.1.1` fixes the issue.

Thanks! :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "escape-string-regexp": "^1.0.2",
     "has-ansi": "^2.0.0",
     "strip-ansi": "^3.0.0",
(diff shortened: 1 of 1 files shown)


Qix-:
I'll let @sindresorhus merge and publish this. :)

sindresorhus:
There's a [new version of Electron out now](https://github.com/atom/electron/releases/tag/v0.32.3), so I suggest we just revert the Electron check in `supports-color` rather than prograting the fix further.

arthurvr:
> so I suggest we just revert the Electron check in supports-color rather than prograting the fix further.

:+1:

PR #85, opened 2015-09-13
New URL for cmder

jorrit:


Qix-:
:+1: Thank you! Good catch.

PR #86, opened 2015-10-08
Add ES2015 template literal example.

### LINKED ISSUE #77: Show example usage with ES2015 template string in readme
Eg https://github.com/studio-b12/keyword-popularity/blob/d82a032af87259248997e6d99c360e634909db54/module/bin/help/examples.js

edvinerikson:
I added two tiny examples, is this enough?
Feedback is appreciated!

Fixes #77

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -68,6 +68,14 @@ chalk.green(
 	chalk.blue.underline.bold('with a blue substring') +
 	' that becomes green again!'
 );
+
+// ES2015 template literals
+const systemStats = `
+CPU: ${chalk.red('90%')}
+RAM: ${chalk.green('40%')}
+DISK: ${chalk.yellow('70%')}
+`;
+
 ```
 
 Easily define your own themes.


readme.md
```diff
@@ -68,6 +68,19 @@ chalk.green(
 	chalk.blue.underline.bold('with a blue substring') +
 	' that becomes green again!'
 );
+
+// ES2015 template literals
+chalk.red(
+	`I am a red line ${calk.blue.underline.bold('with a blue substring')} ` +
```
kevva: `calk` => `chalk`

sindresorhus: Using string concatenation is not a good example of template literals as their main use is not having to do that.

edvinerikson: Indeed. I did that because the line would have been to long otherwise and if I'd continued on a second line I would have ended up with a `\n` (and spaces/tabs if I also use indentation).

I feel that the first example isn't giving you that much, maybe we should go with second only?

sindresorhus: Yeah, I think one example is enough anyways.

edvinerikson:
Alright, I removed the first example. Is this good to go?

sindresorhus:
Yes, thanks :)

PR #88, opened 2015-11-03
Bump dependencies

TrySound:


sindresorhus:
We're not ready to do that yet. The `supports-color` update is going to be updated with some other dependencies when we're ready.

TrySound:
Okay

PR #91, opened 2015-12-12
Remove has-ansi and strip-ansi dependencies

### LINKED ISSUE #70: Remove `has-ansi` and `strip-ansi` dependencies
I never use them and in hindsight it was a mistake to bundle them. They're only useful in some instances and then you can just require them directly.

vdemedes:
This PR fixes #70, removes `has-ansi` and `strip-ansi` dependencies as well as corresponding methods.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,6 @@
 'use strict';
 var escapeStringRegexp = require('escape-string-regexp');
 var ansiStyles = require('ansi-styles');
-var stripAnsi = require('strip-ansi');
-var hasAnsi = require('has-ansi');
 var supportsColor = require('supports-color');
 var defineProps = Object.defineProperties;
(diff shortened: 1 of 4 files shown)


Qix-:
LGTM. Forgot about that ticket, actually. I'd merge it but I want to see what @sindresorhus says regarding versioning.

sindresorhus:
Need to update readme and tests too.

*This will need a major version, so won't be released right away as there are other things we would like to land for it.*

vdemedes:
Tests were updated and just pushed an update to Readme.md. No sight of `has-ansi` and `strip-ansi` now, except in `Related` section of the Readme.

Qix-:
@sindresorhus maybe like chalk/ansi-styles#13 *poke*

sindresorhus:
@Qix- Yes.. 😳

PR #92, opened 2015-12-15
Closing before and reopening the style after a line break

martinheidegger:
In mac terminals line breaks inside a styled text - particularly at the end of a string - can break the color rendering. Take this example:

```JavaScript
chalk.bgGreen('hello\n')
```

This text embedded in another longer paragraph would look like this:

screenshot 2015-12-16 00 19 02

After this PR it would be rendered as below.

screenshot 2015-12-16 00 17 59

Caveats: I havn't tested this on other platforms and the code is not exactly super-high-performing but it should do the job.

_Note: Screenshots taken with iTerm2 on latest Mac OS_

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,6 +83,11 @@ function applyStyle() {
 		// otherwise only the part of the string until said closing code
 		// will be colored, and the rest will simply be 'plain'.
 		str = code.open + str.replace(code.closeRe, code.open) + code.close;
+
+		// Close the coloring before a line break and reopening after next line
+		// To fix a bleed issue on macs
+		// see https://github.com/chalk/chalk/pull/92
+		str = str.replace(/\n/gm, code.close + '\n' + code.open);
 	}
 
 	// Reset the original 'dim' if we changed it to work around the Windows dimmed gray issue.
--- a/test.js
+++ b/test.js
@@ -61,6 +61,10 @@ describe('chalk', function () {
 	it('don\'t output escape codes if the input is empty', function () {
 		assert.equal(chalk.red(), '');
 	});
+
+	it('line breaks should open and close colors', function () {
+		assert.equal(chalk.grey('hello\nworld'), '\u001b[90mhello\u001b[39m\n\u001b[90mworld\u001b[39m');
+	});
 });
 
 describe('chalk on windows', function () {


iancrowther:
nice

Qix-:
Cannot reproduce.

Without Chalk (basic terminal):
screen shot 2015-12-15 at 14 20 34

With Chalk (basic terminal):
screen shot 2015-12-15 at 14 21 56

Without Chalk (iTerm2):
screen shot 2015-12-15 at 14 24 04

With Chalk (iTerm2):
screen shot 2015-12-15 at 14 25 26

---

Wait, nevermind. You can see it with background styles.

This is specific to iTerm 2, and is a bug. It appears iTerm2 is clearing the new line with the existing render mode styles and resetting the cursor back to write any new text, which makes all trailing negative space have the previous line's render mode.

screen shot 2015-12-15 at 14 30 24

This doesn't occur in the default terminal (correctly):

screen shot 2015-12-15 at 14 30 03

If @sindresorhus wants to accept this as a workaround bug, then fine. I don't think it belongs in Chalk though (however, thank you for pointing this out!).

I will file a bug with iTerm2.

Qix-:
Well I wrote up a long-ass comment first about how I couldn't reproduce, then how it was an iTerm2 bug, then I realized I messed up with the default terminal and the default terminal does it too.

Let me research this some more. If this is a bug in both terminals then it needs to be addressed as such and shouldn't belong in Chalk (_technically_).

index.js
```diff
@@ -83,6 +83,11 @@ function applyStyle() {
 		// otherwise only the part of the string until said closing code
 		// will be colored, and the rest will simply be 'plain'.
 		str = code.open + str.replace(code.closeRe, code.open) + code.close;
+
+		// Close the coloring before a line break and reopening after next line
+		// To fix a bleed issue on macs
```
Qix-: Minor nit: lowercase capitalization on `To`.

Qix-:
Also, I want this one to have more tests. Needs to be checked with all of the non-color render escapes as well (bold, reverse, etc.)

Qix-:
screen shot 2015-12-15 at 14 38 02

Ubuntu correctly renders this, as I'd expect. This is incorrect behavior in iTerm2 and OSX Terminal.

Up to @sindresorhus if we want to have a workaround in Chalk. I'm indifferent.

dthree:
I think I've seen this happen in Putty as well.

martinheidegger:
@Qix- Thank you for taking so much effort for it. I am completely overwhelmed by my workload atm. and can not invest more time into this bug for a while. It would be awesome though if someone would be able to pursue this further.

Qix-:
@dthree I'd be interested to see if this is the case.

@martinheidegger no problem :) the owner of iTerm2 has already responded, and from what I can tell this is incorrect functionality, but was chosen to better match XTerm and OSX Terminal originally. I've responded; we'll see what he says.

Qix-:
This actually looks like correct behavior on a second review. George Nachman (the owner of iTerm2) followed both XTerm and VT520 when he made iTerm2, and the VT520 spec states

> 2.9.7 Erase Color
>
> The Erase color selection controls the background color used when text is erased or new text is scrolled on to the screen. Screen background causes newly erased areas or scrolled text to be written using color index zero, the screen background. This is VT and DECterm compatible. Text background causes erased areas or scrolled text to be written using the current text background color. This is PC console compatible and is the factory default.

This means when a new line occurs, the emulator erases the next line with the existing render mode, which makes sense. Since you issue the reset color escape _after_ the new-line, the erase has already occurred, and resetting takes place before text is outputted.

This is _technically_ correct behavior, which means I'm in favor of this PR as this is more user-expected.

@sindresorhus, thoughts?

index.js
```diff
@@ -83,6 +83,11 @@ function applyStyle() {
...
+		// see https://github.com/chalk/chalk/pull/92
+		str = str.replace(/\n/gm, code.close + '\n' + code.open);
```
Qix-: Could you make this `(/\r?\n/gm, code.close + '$0' + code.open)`

Qix-:
This has been deemed correct behavior. The example provided by Apple was vim. If a color scheme has a custom background color, Background Color Erase behaves this way to color the background.

I'm in favor of this PR.

index.js
sindresorhus: @Qix- Tests are not passing with your change. Also, what's the point of using the `m` flag?

Qix-: There isnt. Shouldnt be there.

Qix-: What tests don't pass with the change?

sindresorhus: The one added in this PR.

Qix-: Could have sword `$0` worked. Strange. Anyway updated the original comment with the new call that should be there.

sindresorhus: @Qix- Ok, I went with `(/\r?\n/g, code.close + '$&' + code.open)` instead.

Qix-: :+1: TIL.

index.js
yassine41531, line 88: ........--

PR #95, opened 2016-01-05
tweak defining proto

stevemao:


package.json
```diff
@@ -47,16 +47,16 @@
...
     "coveralls": "^2.11.2",
     "matcha": "^0.6.0",
     "mocha": "*",
-    "nyc": "^3.0.0",
+    "nyc": "^5.2.0",
     "require-uncached": "^1.0.2",
-    "resolve-from": "^1.0.0",
-    "semver": "^4.3.3",
+    "resolve-from": "^2.0.0",
+    "semver": "^5.1.0",
```
sindresorhus: Don't do unrelated changes. Bumping deps should be a separate PR.

sindresorhus:
The description field is there for a reason ;) Could you describe what you've changed and why?

Qix-:
All this PR seems to do is to remove the function that generates the Chalk prototype in lieu of a simple object. There are reasons why we do this. Could you explain the decision behind changing this?

jbnicolai:
The proposed implementation is simpler, and has less code.. so I'm all for :wink: 

@Qix- what's the reason? I suspect it's an artifact of refactor-fu back when trying to make the whole chained-call implementation performant, but perhaps I'm unaware of the slight difference between the current implementation's behaviour and that of the PR.

All tests seem to pass though, and running the benchmark 10 times on both reveals no difference in performance.

```bash
for i in {0..10}; do npm run bench; done | grep Elapsed
```

on `master`:

```
  Elapsed: 5,773.57 ms
  Elapsed: 5,819.39 ms
  Elapsed: 5,974.76 ms
  Elapsed: 5,678.02 ms
  Elapsed: 5,809.12 ms
  Elapsed: 5,790.01 ms
  Elapsed: 5,809.15 ms
  Elapsed: 5,724.53 ms
  Elapsed: 5,870.90 ms
  Elapsed: 5,789.78 ms
  Elapsed: 5,970.74 ms
```

on `stevemao/cleanup`:

```
  Elapsed: 5,770.72 ms
  Elapsed: 5,770.93 ms
  Elapsed: 5,917.21 ms
  Elapsed: 5,872.29 ms
  Elapsed: 5,757.63 ms
  Elapsed: 5,804.15 ms
  Elapsed: 5,733.41 ms
  Elapsed: 5,950.70 ms
  Elapsed: 5,967.84 ms
  Elapsed: 5,961.10 ms
  Elapsed: 5,798.23 ms
```

index.js
```diff
@@ -91,21 +87,7 @@ function applyStyle() {
...
-			get: function () {
-				return build.call(this, [name]);
-			}
-		};
-	});
-
-	return ret;
-}
-
-defineProps(Chalk.prototype, init());
```
stevemao: This is to just make the code simpler. and the logic is pretty much the same as generating the `styles` above so I'm reusing it.

index.js
```diff
@@ -15,23 +15,19 @@ if (isSimpleWindowsTerm) {
...
-})();
+	styles[key] = {
+		get: function () {
+			return build.call(this, this._styles ? this._styles.concat(key) : [key]);
+		}
+	};
+});
 
-var proto = defineProps(function chalk() {}, styles);
+var proto = defineProps({}, styles);
```
stevemao: I'll do a bit more testing on this :smile:

index.js
```diff
@@ -15,23 +15,19 @@ if (isSimpleWindowsTerm) {
 	ansiStyles.blue.open = '\u001b[94m';
 }
 
-var styles = (function () {
-	var ret = {};
```
stevemao: I don't think there's any particular reason to use an iffy here. Removing it to gain a bit (not too much) performance and make the code simpler.

stevemao:
I'll make the changes tonight :)

stevemao:
I broke this into three PRs (#96, #97)

stevemao:
Any questions/problems with this PR?

sindresorhus:
I'm busy at the moment, will try to review in a few days.

@jbnicolai @Qix- Looks good to you?

Qix-:
Yeah this looks fine. I misread what happened with the removal of the function.

sindresorhus:
@stevemao Awesome! Thank you. I dig PRs that simplify the code :)

PR #101, opened 2016-02-29
Use forkeys

Jameskmonger:
Forkeys is an abstraction of the `Object.keys(...).forEach(...)` native feature. It simplifies the code in `index.js` a little.


sindresorhus:
Thanks for the suggestion, but we're not adding more dependencies to `chalk` as this point as it's used sooo much in the wild.

Jameskmonger:
No problem @sindresorhus.

PR #104, opened 2016-03-15
7,700 Dependent Modules

popey456963:
Yet another 1,000 modules in only two months.  That's almost a project an hour starting to use Chalk!


Qix-:
That's pretty crazy. Thanks! :tada:

sindresorhus:
unicorn

PR #116, opened 2016-06-14
Added little patch for compatibility with phantomjs

ket4yii:


sindresorhus:
No thanks. Supporting PhantomJS is a slippery slope. Chalk is for Node.js modules. If you need it elsewhere, browserify or webpack it before use ;)

PR #120, opened 2016-07-23
minor doc improvements

thefourtheye:


readme.md
```diff
@@ -32,7 +32,7 @@
 
 ## Install
 
-```
+```bash
```
sindresorhus: This is not correct as the below snippet is not valid Bash because of the `$`, but you could use `console`.

readme.md
```diff
@@ -157,7 +157,7 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
 - `red`
 - `green`
 - `yellow`
-- `blue` *(on Windows the bright version is used as normal blue is illegible)*
+- `blue` *(on Windows the bright version is used, as normal blue is illegible)*
```
Qix-: The added comma is unnecessary.

thefourtheye: Thanks for the review. Not a native English speaker. It didn't make sense to me without a punctuation mark.

Qix-: Could be `...is used since normal blue...` (`since` instead of `as`). :)

thefourtheye:
@sindresorhus @Qix- Addressed the comments. Thanks for the review :-)

Qix-:
Thank you! :)

PR #129, opened 2016-09-08
Add log statement to chalk examples

### LINKED ISSUE #128: "Here without console.log for purity."
I think this is what is throwing people off. A large portion of our tickets (and things I see on StackOverflow) are people misunderstanding how to use chalk. I think they expect `chalk.red("hello")` to output something.

The readme states `Here without console.log for purity.` which a lot of people miss (even though there's a usage block _right above it_). It might be worthwhile to just include `console.log` (or even `var log = console.log; log(...)`) in there to illustrate that chalk only returns strings with escape sequences, nothing more.

Just a thought.

Qix-:
Added what are essentially `console.log()` calls to the chalk examples.

Closes #128.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,11 +71,11 @@ log(chalk.green(
 ));
 
 // ES2015 template literal
-const systemStats = `
+log(`
 CPU: ${chalk.red('90%')}
 RAM: ${chalk.green('40%')}
 DISK: ${chalk.yellow('70%')}
-`;
+`);
 ```
 
 Easily define your own themes.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -47,29 +47,28 @@ console.log(chalk.blue('Hello world!'));
...
-chalk.green(
+log(chalk.green(
 	'I am a green line ' +
 	chalk.blue.underline.bold('with a blue substring') +
 	' that becomes green again!'
-);
+));
 
 // ES2015 template literal
 const systemStats = `
```
sindresorhus: Can you add a `log` for this one too?

Qix-: Yep :)

PR #136, opened 2016-10-27
Add a new ES6 example using tagged template string

tuhoojabotti:
I found out that you can use chalk functions as template string tags. I thought this would be a nice example to add to the readme.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -76,6 +76,15 @@ CPU: ${chalk.red('90%')}
 RAM: ${chalk.green('40%')}
 DISK: ${chalk.yellow('70%')}
 `);
+
+// ES2015 using [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals)
+const { red, yellow, green } = chalk;
+log(`
+CPU: ${red.bold`90%`}
+RAM: ${green`40%`}
+DISK: ${yellow`70%`}
+`);
+
 ```
 
 Easily define your own themes.


sindresorhus:
What's the point? It's incorrect use of tagged template literals and will not work correctly if the template literal has any variables. You also just save 2 characters.

tuhoojabotti:
The point is just personal style. I think it's the same reason why colors.js used the String prototype, because 'foo'.blue is visually cleaner than blue('foo')

How is it incorrect use? It's true that chalk does not support complicated template strings which does pose a problem for the example because you have to know it, but maybe we could add support to chalk. For example chalk.tagged, which is meant to be used a tag.

I understand if you don't want to merge this example, but if you accept the use case I can try to work on a PR for the tagged-method. Or maybe I'll make my own wrapper module for chalk.

sindresorhus:
This is not about personal style. It's incorrect and unnecessary use of tagged template literals.

I would recommend reading up about its purpose. If we were to add support for it, I think it would be something closer to:

```js
chalk.tpl`foo {chalk.red}bar{chalk.reset} baz`;
```

And I don't really see that being simpler than:

```js
`foo ${chalk.red('bar')} baz`;
```

Maybe for longer colored strings.

Happy to discuss this further when we're on the same page about its purpose though :)

tuhoojabotti:
I know the purpose is to enable DSL building like you show in your example, and while my usage pattern might be against the original purpose of the spec I wouldn't go as far to say that it's incorrect. In the end tagged template literal is just a way to invoke a function with arguments. I understand your opinion on not supporting ``chalk.tag.blue`blue text` `` style in the core library.

Maybe I should be using something like CoffeeScript to get rid of parenthesis. :)

Issue #126, opened 2016-09-01
Switch from ANSI-styles to ANSI-256-colors


ev1stensberg:
https://github.com/jbnicolai/ansi-256-colors

Would it be convenient  to switch to using ansi 256 in order to have more colors provided?

sindresorhus:
[ansi-styles](https://github.com/chalk/ansi-styles) already supports 256 colors and more. We just need to find time to do a new release and use it in Chalk.

ev1stensberg:
Ooh, great. Any expected time? 1mth or so?

ev1stensberg:
FWIW -- Instead of options, perhaps something like `chalk.customconfig.with.lots.of.dots(...)` we could do `chalk.extend()` and let them extend the module ( or similar ) ?

Issue #87, opened 2015-10-09
How to create orange


mikeerickson:
I am looking to display console message in orange, but I am having difficulty finding orange?

sindresorhus:
Currently `chalk` only supports the basic ANSI colors (as those are the ones most terminals supports), which doesn't include orange. We are discussing how to bring in support for 256 colors and true color which will get you orange and lots of other colors: https://github.com/chalk/chalk/issues/73

Qix-:
When #140 is merged, you'll be able to use `chalk.keyword('orange')`.

Issue #35, opened 2014-07-10
8-bit/24-bit color support


rektide:
`\e[38;2;{R};{G};{B}m` is allegedly the ISO-8613-3 standard for 24-bit colors. Would love to have some real color in the terminal. 256-color mode is more complicated, but also would be good to have support for.

Wikipedia is surprisingly helpful on [these ANSI color codes](http://en.wikipedia.org/wiki/ANSI_escape_code#Colors).

rektide:
#73 had some further work on this topic.

Issue #20, opened 2014-05-23
256 color support?


ccheever:
I made a drop-in replacement for chalk that has 256 color support and a few other things.
https://github.com/ccheever/chalk256

Are you interested in adding any of that to chalk? I can create a pull request, but would have to do a bunch of work to make the code match your style, etc., so wanted to ask before putting in the time.

PR #140, opened 2017-01-17
Add RGB (256/Truecolor) support

### LINKED ISSUE #73: Explore a 256 and 'true color' API
See chalk/supports-color#25 and chalk/ansi-styles#11

Once chalk/supports-color and chalk/ansi-styles land support for 256 and truecolor we might want to expose this in chalk as well.

Any suggestions for a simple, clean API?

Qix-:
Finally closes #73.

References #20, #35, #37, #57, #87, #126. We know some of you have been waiting for this one for a very long time.

This PR is blocked by a few remaining issues over at [`supports-color`](https://github.com/chalk/supports-color). @sindresorhus has promised me he will take a look at those and make some executive decisions within the coming week. After they are resolved, its version will be bumped in this PR and, pending approval from the community, this will be merged and released.

This is an additive release - however, please see the breaking changes section below for a reason why it's a major release.

This adds two new suites of chalk-able modes:

1. Foreground RGB colors using [`color-convert`](https://github.com/qix-/color-convert) models as names (e.g. `chalk.rgb(123, 45, 67)` or `chalk.keyword('orange')`)
2. Background RGB colors, similar to the foreground colors above but with a camel-cased `bg` prefix, (e.g. `chalk.bgRgb(123, 45, 67)` or `chalk.bgKeyword('orange')`)

All RGB color functions in turn return your usual chalk builder. Some examples:

- `chalk.hex('#DEADED').underline('Hello, world!')`
- `chalk.keyword('orange')('Some orange text')`
- `chalk.rgb(15, 100, 204).inverse('Hello!')`

This is per [@silverwind's suggestion](https://github.com/chalk/chalk/issues/73#issuecomment-124826052) in #73, which we agreed is the most intuitive way to approach this API.

---

**This module now requires Node.js 4 or above**. This is why it is a major release bump.

The only potentially API-related breaking changes is if you use Chalk's `.enabled` property, you'll need to replace it with `.level`, which is now an integer (where `0` indicates not enabled and anything >= 1 indicates enabled, and further, one of several levels of color granularity).

As well, if you're passing an option to chalk to forcefully set whether or not it's enabled, you'll need to change:

- `{enabled: false}` -> `{level: 0}`, or
- `{enabled: true}` -> `{level: 1}` (or whatever level you'd like - level 1 is equivalent to Chalk's current level of support).

The levels, as described in [`supports-color`](https://github.com/chalk/supports-color), are:

0. Disabled
1. 16 colors (basic color)
2. 256 colors (most widely supported mode in terminals that support more than 16 colors)
3. 16 million colors ("truecolor") - be sure to check [if your emulator supports Truecolor codes](https://gist.github.com/XVilka/8346728)

---

Please don't hesitate to give me feedback on the new API.

Thanks for everyone's continued patience - let this be a celebration of some of the first OSS contributions I made here on Github, and let it serve as a relic of some of the great connections I've made through it. 🦄 

---

cc @sindresorhus @jbnicolai @dthree @silverwind @stevemao @arthurvr @mikeerickson @rektide @mattbasta, as well as @XVilka (Thank you for your research on support! It has [proven invaluable](http://stackoverflow.com/a/26665998/510036) in many ways.)

Thanks everyone!

Look at all the colors!!!!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - 'stable'
   - '6'
   - '4'
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - 'stable'
   - '6'
   - '4'
-  - '0.12'
-  - '0.10'
 after_success: npm run coveralls
--- a/index.js
+++ b/index.js
@@ -6,6 +6,9 @@ var supportsColor = require('supports-color');
 var defineProps = Object.defineProperties;
 var isSimpleWindowsTerm = process.platform === 'win32' && !/^xterm/i.test(process.env.TERM);
 
+var levelMapping = ['ansi', 'ansi', 'ansi256', 'ansi16m'];
+var skipModels = ['gray'];
+
 function Chalk(options) {
 	// detect level if not set manually
 	this.level = !options || options.level === undefined ? supportsColor.level : options.level;
@@ -25,6 +28,43 @@ Object.keys(ansiStyles).forEach(function (key) {
 		get: function () {
 			var codes = ansiStyles[key];
 			return build.call(this, this._styles ? this._styles.concat(codes) : [codes], key);
+		}
+	};
+});
+
+ansiStyles.color.closeRe = new RegExp(escapeStringRegexp(ansiStyles.color.close), 'g');
+Object.keys(ansiStyles.color.ansi).forEach(function (model) {
+	if (skipModels.indexOf(model) !== -1) {
+		return;
+	}
+
+	styles[model] = {
+		get: function () {
+			var level = this.level;
+			return function () {
+				var open = ansiStyles.color[levelMapping[level]][model].apply(null, arguments);
+				var codes = {open: open, close: ansiStyles.color.close, closeRe: ansiStyles.color.closeRe};
+				return build.call(this, this._styles ? this._styles.concat(codes) : [codes], model);
+			};
+		}
+	};
+});
+
+ansiStyles.bgColor.closeRe = new RegExp(escapeStringRegexp(ansiStyles.bgColor.close), 'g');
+Object.keys(ansiStyles.bgColor.ansi).forEach(function (model) {
+	if (skipModels.indexOf(model) !== -1) {
+		return;
+	}
+
+	var bgModel = 'bg' + model.charAt(0).toUpperCase() + model.substring(1);
+	styles[bgModel] = {
+		get: function () {
+			var level = this.level;
+			return function () {
+				var open = ansiStyles.bgColor[levelMapping[level]][model].apply(null, arguments);
+				var codes = {open: open, close: ansiStyles.bgColor.close, closeRe: ansiStyles.bgColor.closeRe};
+				return build.call(this, this._styles ? this._styles.concat(codes) : [codes], model);
+			};
 		}
 	};
 });
--- a/index.js
+++ b/index.js
@@ -6,7 +6,9 @@ var supportsColor = require('supports-color');
 var defineProps = Object.defineProperties;
 var isSimpleWindowsTerm = process.platform === 'win32' && !/^xterm/i.test(process.env.TERM);
 
+// supportsColor.level -> ansiStyles.color[name] mapping
 var levelMapping = ['ansi', 'ansi', 'ansi256', 'ansi16m'];
+// color-convert models to exclude from the Chalk API due to conflicts and such.
 var skipModels = ['gray'];
 
 function Chalk(options) {
--- a/index.js
+++ b/index.js
@@ -7,8 +7,8 @@ var defineProps = Object.defineProperties;
 var isSimpleWindowsTerm = process.platform === 'win32' && !/^xterm/i.test(process.env.TERM);
 
 function Chalk(options) {
-	// detect mode if not set manually
-	this.enabled = !options || options.enabled === undefined ? supportsColor : options.enabled;
+	// detect level if not set manually
+	this.level = !options || options.level === undefined ? supportsColor.level : options.level;
 }
 
 // use bright blue on Windows as the normal blue color is illegible
@@ -37,7 +37,7 @@ function build(_styles) {
 	};
 
 	builder._styles = _styles;
-	builder.enabled = this.enabled;
+	builder.level = this.level;
 	// __proto__ is used because we must return a function, but there is
 	// no way to create a function with a different prototype.
 	/* eslint-disable no-proto */
@@ -59,7 +59,7 @@ function applyStyle() {
 		}
 	}
 
-	if (!this.enabled || !str) {
+	if (!this.level || !str) {
 		return str;
 	}
 
--- a/index.js
+++ b/index.js
@@ -82,7 +82,7 @@ function build(_styles, key) {
 	var self = this;
 
 	builder._styles = _styles;
-  
+
 	Object.defineProperty(builder, 'level', {
 		enumerable: true,
 		get: function () {
--- a/index.js
+++ b/index.js
@@ -23,7 +23,8 @@ Object.keys(ansiStyles).forEach(function (key) {
 
 	styles[key] = {
 		get: function () {
-			return build.call(this, this._styles ? this._styles.concat(key) : [key]);
+			var codes = ansiStyles[key];
+			return build.call(this, this._styles ? this._styles.concat(codes) : [codes], key);
 		}
 	};
 });
@@ -31,13 +32,17 @@ Object.keys(ansiStyles).forEach(function (key) {
 // eslint-disable-next-line func-names
 var proto = defineProps(function chalk() {}, styles);
 
-function build(_styles) {
+function build(_styles, key) {
 	var builder = function () {
 		return applyStyle.apply(builder, arguments);
 	};
 
 	builder._styles = _styles;
 	builder.level = this.level;
+
+	// see below for fix regarding invisible grey/dim combination on windows.
+	builder.hasGrey = this.hasGrey || key === 'gray' || key === 'grey';
+
 	// __proto__ is used because we must return a function, but there is
 	// no way to create a function with a different prototype.
 	/* eslint-disable no-proto */
@@ -70,12 +75,12 @@ function applyStyle() {
 	// see https://github.com/chalk/chalk/issues/58
 	// If we're on Windows and we're dealing with a gray color, temporarily make 'dim' a noop.
 	var originalDim = ansiStyles.dim.open;
-	if (isSimpleWindowsTerm && (nestedStyles.indexOf('gray') !== -1 || nestedStyles.indexOf('grey') !== -1)) {
+	if (isSimpleWindowsTerm && this.hasGrey) {
 		ansiStyles.dim.open = '';
 	}
 
 	while (i--) {
-		var code = ansiStyles[nestedStyles[i]];
+		var code = nestedStyles[i];
 
 		// Replace any instances already present with a re-opening code
 		// otherwise only the part of the string until said closing code
--- a/readme.md
+++ b/readme.md
@@ -212,6 +212,8 @@ For a complete list of color models, see [`color-convert`'s list of conversions]
 As of this writing, these are the supported color models that are exposed in Chalk:
 
 - `rgb`
+- `hex`
+- `keyword` (CSS keywords)
 - `hsl`
 - `hsv`
 - `hwb`
@@ -219,12 +221,10 @@ As of this writing, these are the supported color models that are exposed in Cha
 - `xyz`
 - `lab`
 - `lch`
-- `hex`
-- `keyword`
 - `ansi16`
 - `ansi256`
 - `hcg`
-- `apple`
+- `apple` (see [qix-/color-convert#30](https://github.com/Qix-/color-convert/issues/30))
 
 ## Windows
 
--- a/readme.md
+++ b/readme.md
@@ -126,7 +126,7 @@ const ctx = new chalk.constructor({level: 0});
 
 Levels are as follows:
 
-0. All color disabled
+0. All colors disabled
 1. Basic color support (16 colors)
 2. 256 color support
 3. RGB/Truecolor support (16 million colors)
--- a/readme.md
+++ b/readme.md
@@ -198,12 +198,11 @@ For supported methods (listed below), the color will be 'fit' to the color level
 
 Some examples:
 
-For a complete list of color models, see [`color-convert`'s list of conversions](https://github.com/Qix-/color-convert/blob/master/conversions.js). Background versions of these models are prefixed with `bg` and the first level of the module capitalized (e.g. `keyword` for foreground colors and `bgKeyword` for background colors).
-
 - `chalk.hex('#DEADED').underline('Hello, world!')`
 - `chalk.keyword('orange')('Some orange text')`
 - `chalk.rgb(15, 100, 204).inverse('Hello!')`
 
+Background versions of these models are prefixed with `bg` and the first level of the module capitalized (e.g. `keyword` for foreground colors and `bgKeyword` for background colors).
 
 - `chalk.bgHex('#DEADED').underline('Hello, world!')`
 - `chalk.bgKeyword('orange')('Some orange text')`
@@ -225,6 +224,8 @@ As of this writing, these are the supported color models that are exposed in Cha
 - `ansi256`
 - `hcg`
 - `apple` (see [qix-/color-convert#30](https://github.com/Qix-/color-convert/issues/30))
+
+For a complete list of color models, see [`color-convert`'s list of conversions](https://github.com/Qix-/color-convert/blob/master/conversions.js).
 
 ## Windows
 
--- a/readme.md
+++ b/readme.md
@@ -76,14 +76,23 @@ CPU: ${chalk.red('90%')}
 RAM: ${chalk.green('40%')}
 DISK: ${chalk.yellow('70%')}
 `);
+
+// Use RGB colors in terminal emulators that support it.
+log(chalk.keyword('orange')('Yay for orange colored text!'));
+log(chalk.rgb(123, 45, 67).underline('Underlined reddish color'));
+log(chalk.hex('#DEADED').bold('Bold gray!'));
 ```
 
 Easily define your own themes.
 
 ```js
 const chalk = require('chalk');
+
 const error = chalk.bold.red;
+const warning = chalk.keyword('orange');
+
 console.log(error('Error!'));
+console.log(warning('Warning!'));
 ```
 
 Take advantage of console.log [string substitution](http://nodejs.org/docs/latest/api/console.html#console_console_log_data).
@@ -105,15 +114,22 @@ Chain [styles](#styles) and call the last one as a method with a string argument
 
 Multiple arguments will be separated by space.
 
-### chalk.enabled
+### chalk.level
 
-Color support is automatically detected, but you can override it by setting the `enabled` property. You should however only do this in your own code as it applies globally to all chalk consumers.
+Color support is automatically detected, but you can override it by setting the `level` property. You should however only do this in your own code as it applies globally to all chalk consumers.
 
 If you need to change this in a reusable module create a new instance:
 
 ```js
-const ctx = new chalk.constructor({enabled: false});
+const ctx = new chalk.constructor({level: 0});
 ```
+
+Levels are as follows:
+
+0. All color disabled
+1. Basic color support (16 colors)
+2. 256 color support
+3. RGB/Truecolor support (16 million colors)
 
 ### chalk.supportsColor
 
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -1,8 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - 'stable'
```
sindresorhus: This is not needed. We already define 6 and 4.

index.js
```diff
@@ -6,9 +6,12 @@ var supportsColor = require('supports-color');
 var defineProps = Object.defineProperties;
 var isSimpleWindowsTerm = process.platform === 'win32' && !/^xterm/i.test(process.env.TERM);
 
+var levelMapping = ['ansi', 'ansi', 'ansi256', 'ansi16m'];
```
sindresorhus: You have `ansi` twice there.

index.js
```diff
@@ -6,9 +6,12 @@ var supportsColor = require('supports-color');
...
+var skipModels = ['gray'];
```
sindresorhus: Did you mean `skipModes`?

index.js
```diff
@@ -6,9 +6,12 @@ var supportsColor = require('supports-color');
...
+
 function Chalk(options) {
-	// detect mode if not set manually
-	this.enabled = !options || options.enabled === undefined ? supportsColor : options.enabled;
+	// detect level if not set manually
+	this.level = !options || options.level === undefined ? supportsColor.level : options.level;
```
sindresorhus: I think we should still support `enabled`. Not enough benefit to make such a big breaking change. We can support both. Most will only need to know whether it's enabled or not anyways.

index.js
sindresorhus: Since we target Node.js 4 now, we can use `Set` on this and the below. Will be faster.

index.js
```diff
@@ -23,21 +26,63 @@ Object.keys(ansiStyles).forEach(function (key) {
...
 		get: function () {
-			return build.call(this, this._styles ? this._styles.concat(key) : [key]);
+			var codes = ansiStyles[key];
+			return build.call(this, this._styles ? this._styles.concat(codes) : [codes], key);
+		}
+	};
+});
+
+ansiStyles.color.closeRe = new RegExp(escapeStringRegexp(ansiStyles.color.close), 'g');
+Object.keys(ansiStyles.color.ansi).forEach(function (model) {
```
sindresorhus: Use a for-of loop.

index.js
```diff
@@ -23,21 +26,63 @@ Object.keys(ansiStyles).forEach(function (key) {
...
+				var open = ansiStyles.color[levelMapping[level]][model].apply(null, arguments);
+				var codes = {open: open, close: ansiStyles.color.close, closeRe: ansiStyles.color.closeRe};
+				return build.call(this, this._styles ? this._styles.concat(codes) : [codes], model);
+			};
+		}
+	};
+});
+
+ansiStyles.bgColor.closeRe = new RegExp(escapeStringRegexp(ansiStyles.bgColor.close), 'g');
+Object.keys(ansiStyles.bgColor.ansi).forEach(function (model) {
```
sindresorhus: Use a for-of loop.

readme.md
```diff
@@ -174,10 +190,30 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
 - `bgWhite`
 
 
-## 256-colors
+## 256/16 million (Truecolor) color support
+
+Chalk supports 256 colors and, when manually specified, [Truecolor (16 million colors)](https://gist.github.com/XVilka/8346728) on all supported terminal emulators.
```
sindresorhus: No need for comma after `and`

readme.md
```diff
@@ -174,10 +190,30 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
...
+- `cmyk`
+- `xyz`
+- `lab`
+- `lch`
+- `hex`
+- `keyword`
+- `ansi16`
+- `ansi256`
+- `hcg`
+- `apple`
```
sindresorhus: These needs to be explained, and put the most common ones at the top, like `ansi*` and `hex`.

readme.md
```diff
@@ -174,10 +190,30 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
...
 
 
-## 256-colors
+## 256/16 million (Truecolor) color support
+
+Chalk supports 256 colors and, when manually specified, [Truecolor (16 million colors)](https://gist.github.com/XVilka/8346728) on all supported terminal emulators.
+
+For the methods that support it (listed below), the color will be 'fit' to the color level supported (i.e. RGB colors will be downsampled to 16 colors if only basic support is enabled).
+
+For a complete list of color models, see [`color-convert`'s list of conversions](https://github.com/Qix-/color-convert/blob/master/conversions.js). Background versions of these models are prefixed with `bg` and the first level of the module capitalized (e.g. `keyword` for foreground colors and `bgKeyword` for background colors).
```
sindresorhus: Add some examples.

readme.md
```diff
@@ -105,16 +114,23 @@ Chain [styles](#styles) and call the last one as a method with a string argument
...
 If you need to change this in a reusable module create a new instance:
 
 ```js
-const ctx = new chalk.constructor({enabled: false});
+const ctx = new chalk.constructor({level: 0});
 ```
 
+Levels are as follows:
+
+0. All color disabled
```
sindresorhus: colors

index.js
```diff
@@ -23,21 +26,63 @@ Object.keys(ansiStyles).forEach(function (key) {
...
 	styles[key] = {
 		get: function () {
-			return build.call(this, this._styles ? this._styles.concat(key) : [key]);
+			var codes = ansiStyles[key];
+			return build.call(this, this._styles ? this._styles.concat(codes) : [codes], key);
+		}
+	};
+});
+
+ansiStyles.color.closeRe = new RegExp(escapeStringRegexp(ansiStyles.color.close), 'g');
```
sindresorhus: Why this?

index.js
Qix-: The double-ansi is on purpose - if level is `0` we still need to be pulling _something_ from ansi-styles.

Qix-: Open to suggestions as to how to better approach it. I agree it feels a little clunky. A lot of what chalk does is clunky, though.

index.js
Qix-: No - "model" refers to the color models within color-convert. `gray` was for CSS `gray()` and conflicts with our color `gray` or `grey`.

index.js
Qix-: How do we handle when someone supplies `{level: 0, enabled: true}` or `{level: 2, enabled: false}`?

index.js
Qix-: Why not that? Chalk requires a closing code regex for building the style.

index.js
Qix-: Should we change the original loop as well?

index.js
Qix-: See https://github.com/chalk/chalk/pull/140#discussion_r96395981.

readme.md
Qix-: Why not? It's a qualifier for the subsequent clause. Grammatically the lack of a comma wouldn't make sense here.

index.js
sindresorhus: How you handle mutually exclusive things in general, you throw a user-friendly error, and document it well.

index.js
sindresorhus: I mean, it's a little bit dangerous to assign to `ansiStyles` as it's a single-ton. So if someone uses `ansi-styles` another place, their instance will also be monkey-patched. Might be a non-issue in practise, but wanted to bring it up.

index.js
sindresorhus: Yes, but can do unrelated changes later so not to make this PR too noisy.

readme.md
Qix-: `ansi*` aren't very common since Chalk handles them already. I'll add `hex` to the top.

index.js
sindresorhus: Can you add a short comment explaining that. Was not clear when reading the code.

index.js
sindresorhus: I see. A short comment would solve it.

sindresorhus: Nvm the `Set` comment. I didn't read the code properly.

index.js
Qix-: :+1:

index.js
Qix-: I'm writing this code out and it just doesn't make sense. The code is super hairy - having to check to see if keys are in the configuration object and then do some arbitrary logic on the values. What does `enabled: true` mean now? A straight-up level of `1` or do we use `supportsColor.level`?

We're already bumping the major version and those advanced users using forced enable/disable outside of the functionality supplied by `supports-color` will be willing to adjust how they want Chalk to behave.

index.js
Qix-: Chalk did it before in the loop above. We can change that, though - I'm not opposed.

Qix-: Also, [it's not applied globally](https://github.com/chalk/ansi-styles/blob/master/index.js#L119-L122). We fixed that a while back :)

index.js
Qix-: I think we should do a complete es2015-ification PR as a separate step to prevent history fragmentation, IMO. But I agree that it should happen.

index.js
Qix-: Done :)

readme.md
```diff
@@ -174,10 +190,41 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
...
+- `hsv`
+- `hwb`
+- `cmyk`
+- `xyz`
+- `lab`
+- `lch`
+- `ansi16`
+- `ansi256`
+- `hcg`
+- `apple` (see [qix-/color-convert#30](https://github.com/Qix-/color-convert/issues/30))
```
silverwind: I think one example per color model would be very helpful. Show how a single color is represented through all the various methods. Maybe through a table that contains `method | model name | example code`.

sindresorhus: Most of these color models are not very practically useful though. I doubt most of these will be used by users. I would prefer to only document the most popular ones and rather leave it to color-convert to document the rest.

Qix-: The other thing is that the way color-convert is versioned all new models that are added 'just work' with whatever is using them assuming you're using `^`. This would be an ever changing list.

silverwind: I'm fine with just documenting the most popular ones. I think these are rgb,hex,keyword,hsl,cymk. We can link to the color-convert docs for further reading.

sindresorhus: @silverwind 👍 Except for `cmyk`. I don't see why anyone would use that in the terminal. CMYK is for printing.

readme.md
```diff
@@ -174,10 +190,41 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
 - `bgWhite`
 
 
-## 256-colors
+## 256/16 million (Truecolor) color support
+
+Chalk supports 256 colors and, when manually specified, [Truecolor (16 million colors)](https://gist.github.com/XVilka/8346728) on all supported terminal emulators.
+
+For the methods that support it (listed below), the color will be 'fit' to the color level supported (i.e. RGB colors will be downsampled to 16 colors if only basic support is enabled).
```
silverwind: "For supported methods"

index.js
vadimdemedes: I'd vote on leaving `enabled` as well, mostly for backwards compatibility reasons.

> A straight-up level of 1 or do we use supportsColor.level?

This seems like a good solution. Anything potentially bad with it?

readme.md
Qix-: 👍

readme.md
Qix-: So what should be put there, then? Just a basic description? Or a line of sample code?

index.js
Qix-: Because `enabled` is on by default, so it's even more weird code to determine if we default to `level = 1` or `level = supportsColor.level` based on if something is passed into the Chalk constructor.

index.js
```diff
@@ -23,21 +28,63 @@ Object.keys(ansiStyles).forEach(function (key) {
...
+Object.keys(ansiStyles.color.ansi).forEach(function (model) {
+	if (skipModels.indexOf(model) !== -1) {
```
vadimdemedes: `skipModels.indexOf(model) >= 0` would look nicer, imo.

index.js
```diff
@@ -23,21 +28,63 @@ Object.keys(ansiStyles).forEach(function (key) {
...
+				var codes = {open: open, close: ansiStyles.color.close, closeRe: ansiStyles.color.closeRe};
+				return build.call(this, this._styles ? this._styles.concat(codes) : [codes], model);
+			};
+		}
+	};
+});
+
+ansiStyles.bgColor.closeRe = new RegExp(escapeStringRegexp(ansiStyles.bgColor.close), 'g');
+Object.keys(ansiStyles.bgColor.ansi).forEach(function (model) {
+	if (skipModels.indexOf(model) !== -1) {
```
vadimdemedes: Same as https://github.com/chalk/chalk/pull/140/files#r96501587.

index.js
Qix-: Like...

```javascript
options = options || {};
if ('enabled' in options) {
    this.enabled = options.enabled;
    this.level = options.enabled ? 1 : 0;
} else {
    var level = 'level' in options ? options.level : supportsColor.level;
    this.enabled = +Boolean(supportsColor);
    this.level = level;
}
```

readme.md
```diff
@@ -174,10 +190,41 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
 - `bgWhite`
 
 
-## 256-colors
+## 256/16 million (Truecolor) color support
+
+Chalk supports 256 colors and, when manually specified, [Truecolor (16 million colors)](https://gist.github.com/XVilka/8346728) on all supported terminal emulators.
+
+For supported methods (listed below), the color will be 'fit' to the color level supported (i.e. RGB colors will be downsampled to 16 colors if only basic support is enabled).
```
vadimdemedes: perhaps `will be 'fit'` => `will be converted`?

Qix-: I think `downsampled` was the word I was looking for here.

E.g. if your support level is 2 (256 colors) and you give it an RGB, it will downsample the full 16 million color down to the 256-palette color.

Is `downsample` correct here? I want to be a bit more specific than `convert` since in this context it could be mistaken as converting to another color model (which is technically true, I suppose...)

readme.md
```diff
@@ -174,10 +190,41 @@ console.log(chalk.styles.red.open + 'Hello' + chalk.styles.red.close);
...
+
+Some examples:
+
+For a complete list of color models, see [`color-convert`'s list of conversions](https://github.com/Qix-/color-convert/blob/master/conversions.js). Background versions of these models are prefixed with `bg` and the first level of the module capitalized (e.g. `keyword` for foreground colors and `bgKeyword` for background colors).
```
silverwind: I'd move this paragraph below  the examples.

index.js
Qix-: I tend to err on `!== -1` since the spec specifically states

> The first index of the element in the array; -1 if not found.

It's obviously bike shedding. It'd be nice if XO made a definitive decision on it.

readme.md
Qix-: Agreed. Does that look better?

readme.md
silverwind: Expand the examples to include above 4 using the same color. After that, I guess some work is needed on `color-convert`'s readme to have more examples for the rarely used models. After that, link to that readme (instead of linking to the code).

Qix-: Yeah, I need to start enforcing readme updates in `color-convert`.

Can do.

readme.md
silverwind: Yes, that's fine.

readme.md
vadimdemedes: Yep, this is better, but `downsampled` is in the parenthesis as well. Thinking whether it'd read a bit repetitive.

readme.md
Qix-: How's that?

readme.md
Qix-: > Colors are downsampled from 16 million RGB values to an ANSI color format that is supported by the terminal emulator (or by specifying `{level: n}` as a chalk option). For example, Chalk configured to run at level 1 (basic color support) will downsample an RGB value of `#FF0000` (red) to `31` (ANSI escape for red).

Is that closer? It's such awkward verbiage :|

vadimdemedes:
Great work, @Qix-!

Qix-:
Thank you -- and thank you for all of the valuable feedback, as always!

readme.md
vadimdemedes: Oh yes, I think this is better and more detailed ;)

index.js
vadimdemedes: True that.

readme.md
Qix-: Cool, thanks :D

index.js
vadimdemedes: The code you posted is fine, but I think `enabled` needs to be defined via `Object.defineProperty` and modify/read `this.level` under the hood to support that:

```
chalk.enabled = false;
```

Please correct me if that's wrong, I may be missing things.

vadimdemedes: > I'm genuinely curious as to how many people actually use the custom constructor?

I'm guessing like you do, but I doubt many people use that. To say the truth, I discovered it just now.

Qix-: I personally think the code I posted is bloated and huge just to support a property that doesn't really make sense now that there are different levels of support instead of a boolean on/off value.

If we were to do this, then I agree; `.enabled` would have to reflect the `.level` given a getter/setter descriptor.

readme.md
silverwind: Examples looking good. I guess it's cleaner to remove the models without examples once color-convert's has examples for those.

index.js
vadimdemedes: I'm only voting for `enabled` prop, because I see it somewhat often used in tests to conveniently enable/disable colors when testing output.

Let's see what other guys think.

mikeerickson:
@Qix- Any eta when this will be merged?  Or, is there a branch/tag which I could start using right now?  I have been waiting for this for quite some time :-)

index.js
Qix-: The level prop will have the same functionality, though :) haha

Qix-:
@mikeerickson `npm i chalk/chalk#256-truecolor` :)

mikeerickson:
@Qix- Fantastic!

image

mikeerickson:
@Qix- There seems to be an issue with ansi-styles (webpack UglifyJs throws an issue)

Here is what I am seeing (see the malformed ES6 arrow function)

`ansi-styles/index.js` has a number of these errors

```
const wrapAnsi16 = (fn, offset) => function () {
	const code = fn.apply(colorConvert, arguments);
	return `\u001B[${code + offset}m`;
};
```

mikeerickson:
@Qix- Seems as though this issue was introduced in `ansi-styles 3.0` 

If I revert back to what is installed by `chalk 1.1.3` it will use `ansi-styles 2.2.1`

Qix-:
It's not malformed; it's a lambda that returns a lambda. Not sure when that got changed but it's valid. What's the error message? Note that you need Node >= 4 or you need to babel it.

mikeerickson:
I am using Node 6.8.1 
Webpack fails whe Uglify ( -p ) throwing error at invalid token ( > ) on the ansi16 call outlined above

If I completely remove chalk (which removes ansi-styles) and reinstall chalk 1.1.3 

Then install the version you outlined, all works.

However, if I then remove node_modules and reinstall, ansi-styles 3.0 is installed and webpack again fails

mikeerickson:
@qix sorry to be sort of sparse, replying from phone

mikeerickson:
@qix and I am using babel (via babel-loader in webpack)

Qix-:
What exactly is the error message being produced? And what can I do to reproduce it on my end?

mikeerickson:
@qix I will try and create a small reproducible example.

mikeerickson:
@Qix- OK, I have created a reproducible archive and will post the repo shortly.

mikeerickson:
@Qix- OK, here you go.  Here is a repo which outlines the issue

https://github.com/mikeerickson/chalk-test

Qix-:
@mikeerickson Thank you very much - I can reproduce. It's an UglifyJS bug where it doesn't quite fully support ES6 yet. Check out mishoo/UglifyJS2#448.

Qix-:
@mikeerickson Run this:

```shell
rm -rf node_modules
npm i
find node_modules -type d | grep uglify-js | xargs rm -rf
patch ./node_modules/webpack/package.json <<EOF
70c70
<     "uglify-js": "~2.7.3",
---
>     "uglify-js": "mishoo/UglifyJS2#harmony",
EOF
npm i
npm i chalk/chalk#256-truecolor
node ./index.js
npm run build
```

silverwind:
What a shame that we still don't have a ES6 minifier at this day and age 😢.

sindresorhus:
@silverwind https://github.com/babel/babili

caesarsol:
Hello, this is wonderful, but is the PR abandoned?

Or do you need any help in bringing it further?

doughsay:
I'm looking forward to this one too; is this still in the works?

Qix-:
@sindresorhus ping?

Also, +1 for `babili` - been using it pretty religiously, and while it's quite slow, it's effective and does a perfect job.

sindresorhus:
Yay! At last. Thanks for working on this @Qix-. Sorry about the delay everyone.

1ab7e29c-117e-11e6-9cbb-f0828bacea8e

kevva:
🎉  Great work @Qix-

Qix-:
<3 Thanks everyone

PR #153, opened 2017-05-28
chore(package): bump dependencies

danielsogl:
I updated the dependencies decided to.the latest feature and bugfix releases without breaking changes.

I'm not sure why chalk it self don't use the latest releases of your own packages.


Qix-:
Because we're not ready for them yet.

PR #163, opened 2017-06-29
Add tagged template literal

Qix-:
Got kind of bored tonight and added chalk tagged literals.

Example:

```javascript

console.log(chalk`
CPU: {red.bold.inverse 90%}
RAM: {green.bold 40%}
DISK: {rgb(255,131,0).bold 70%}
`);
```

Gives you:

screen shot 2017-06-28 at 10 18 39 pm

Should be 100% backwards compatible&ast;. Only caveat is that it doesn't work with `new chalk.constructor(options)`. Just have to drop the `new` to get the tag-able version. This, however, means that all existing code that calls `new chalk.constructor` is unaffected.

We might want to update `chalk-cli`, too, as this is a modified implementation to allow for the new 256/truecolor methods.

&ast; Backwards compatible API-wise (down to the `new chalk.constructor(opts)` bit). However unless we want to do `arguments` concatenation acrobatics instead of splats in `index.js`, Node 6 is required :P Up to you, @sindresorhus. I have no strong opinions either way.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,8 @@ const escapeStringRegexp = require('escape-string-regexp');
 const ansiStyles = require('ansi-styles');
 const supportsColor = require('supports-color');
 
+const template = require('./templates.js');
+
 const isSimpleWindowsTerm = process.platform === 'win32' && !process.env.TERM.toLowerCase().startsWith('xterm');
 
 // `supportsColor.level` → `ansiStyles.color[name]` mapping
@@ -11,18 +13,43 @@ const levelMapping = ['ansi', 'ansi', 'ansi256', 'ansi16m'];
 // `color-convert` models to exclude from the Chalk API due to conflicts and such
 const skipModels = new Set(['gray']);
 
-function Chalk(options) {
+const styles = Object.create(null);
+
+function applyOptions(obj, options) {
+	options = options || {};
+
 	// Detect level if not set manually
-	this.level = Number(!options || options.level === undefined ? supportsColor.level : options.level);
-	this.enabled = options && 'enabled' in options ? options.enabled : this.level > 0;
+	obj.level = options.level === undefined ? supportsColor.level : options.level;
+	obj.enabled = 'enabled' in options ? options.enabled : obj.level > 0;
+}
+
+function Chalk(options) {
+	// We check for this.template here since calling chalk.constructor()
+	// by itself will have a `this` of a previously constructed chalk object.
+	if (!this || !(this instanceof Chalk) || this.template) {
+		const chalk = {};
+		applyOptions(chalk, options);
+
+		chalk.template = function () {
+			const args = [].slice.call(arguments);
+			return chalkTag.apply(null, [chalk.template].concat(args));
+		};
+
+		Object.setPrototypeOf(chalk, Chalk.prototype);
+		Object.setPrototypeOf(chalk.template, chalk);
+
+		chalk.template.constructor = Chalk;
+
+		return chalk.template;
+	}
+
+	applyOptions(this, options);
 }
 
 // Use bright blue on Windows as the normal blue color is illegible
 if (isSimpleWindowsTerm) {
 	ansiStyles.blue.open = '\u001B[94m';
 }
-
-const styles = Object.create(null);
 
 for (const key of Object.keys(ansiStyles)) {
 	ansiStyles[key].closeRe = new RegExp(escapeStringRegexp(ansiStyles[key].close), 'g');
@@ -164,7 +191,24 @@ function applyStyle() {
 	return str;
 }
 
+function chalkTag(chalk, strings) {
+	const args = [].slice.call(arguments, 2);
+
+	if (!Array.isArray(strings)) {
+		return strings.toString();
+	}
+
+	const parts = [strings.raw[0]];
+
+	for (let i = 1; i < strings.length; i++) {
+		parts.push(args[i - 1].toString().replace(/[{}]/g, '\\$&'));
+		parts.push(strings.raw[i]);
+	}
+
+	return template(chalk, parts.join(''));
+}
+
 Object.defineProperties(Chalk.prototype, styles);
 
-module.exports = new Chalk();
+module.exports = Chalk(); // eslint-disable-line new-cap
 module.exports.supportsColor = supportsColor;
--- a/templates.js
+++ b/templates.js
@@ -1,0 +1,175 @@
+'use strict';
+
+function data(parent) {
+	return {
+		styles: [],
+		parent,
+		contents: []
+	};
+}
+
+const zeroBound = n => n < 0 ? 0 : n;
+const lastIndex = a => zeroBound(a.length - 1);
+
+const last = a => a[lastIndex(a)];
+
+const takeWhileReverse = (array, predicate, start) => {
+	const out = [];
+
+	for (let i = start; i >= 0 && i <= start; i--) {
+		if (predicate(array[i])) {
+			out.unshift(array[i]);
+		} else {
+			break;
+		}
+	}
+
+	return out;
+};
+
+/**
+ * Checks if the character at position i in string is a normal character a.k.a a non control character.
+ * */
+const isNormalCharacter = (string, i) => {
+	const char = string[i];
+	const backslash = '\\';
+
+	if (!(char === backslash || char === '{' || char === '}')) {
+		return true;
+	}
+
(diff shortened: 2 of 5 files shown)


Qix-:
Wat how did _any_ of these tests work before? These have nothing to do with the changes I made.

Qix-:
Ignore the circleci tests. Travis had a cow recently and required that our tests be updated.

sindresorhus:
> However unless we want to do `arguments` concatenation acrobatics instead of splats in `index.js`, Node 6 is required :P Up to you, [@sindresorhus](https://github.com/sindresorhus). I have no strong opinions either way.

We should support Node.js 4 out its LTS period, as we have so many users.

sindresorhus:
This is really cool. I could have used this when styling long strings.

index.js
```diff
@@ -152,7 +178,18 @@ function applyStyle() {
 	return str;
 }
 
+function chalkTag(chalk, strings, ...args) {
+	const parts = [strings.raw[0]];
+
+	for (let i = 1; i < strings.length; i++) {
+		parts.push(args[i - 1].toString().replace(/([{}])/g, '\\$1'));
```
sindresorhus: `.replace(/([{}])/g, '\\$1')` => `.replace(/[{}]/g, '\\$&')`

sindresorhus: I also couldn't find a test for escaping the arg.

templates.js
```diff
@@ -0,0 +1,172 @@
...
+ * Computes the style for a given data based on it's style and the style of it's parent. Also accounts for !style styles
+ * which remove a style from the list if present.
+ * */
+const sumStyles = data => {
+	const negateRegex = /^~.+/;
+	let out = [];
+
+	for (const style of collectStyles(data)) {
+		if (negateRegex.test(style)) {
+			out = exclude(out, style.slice(1));
```
sindresorhus: Can simply be:

```js
const exclude = style.slice(1);
out = out.filter(x => x !== exclude);
```

And we can drop the `arr-exclude` dependency.

sindresorhus:
Code coverage:

```
--------------|----------|----------|----------|----------|----------------|
File          |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
--------------|----------|----------|----------|----------|----------------|
All files     |    98.19 |    95.12 |     96.3 |    98.11 |                |
 index.js     |      100 |      100 |       90 |      100 |                |
 templates.js |    96.39 |    90.91 |      100 |    96.05 |     19,111,140 |
--------------|----------|----------|----------|----------|----------------|
```

Can you cover those last template.js lines?

sindresorhus:
What's the point of using a tagged template literal though? You're not using it for anything. A function would work just as well.

templates.js
Qix-: :+1: I copied this from `chalk-cli` and did enough to get it to work, but this would be great. Will do.

Qix-:
> What's the point of using a tagged template literal though? You're not using it for anything. A function would work just as well.

```javascript
const logo = '𝚫';

// Because you go from this...

console.log(`
  ${chalk.bold(`${logo} now dns ls`)} [domain]
  ${chalk.bold(
    `${logo} now dns add`
  )} <domain> <name> <A | AAAA | ALIAS | CNAME | TXT> <value>
  ${chalk.bold(`${logo} now dns add`)} <domain> <name> MX <value> <mx_priority>
  ${chalk.bold(
    `${logo} now dns add`
  )} <domain> <name> SRV <priority> <weight> <port> <target>
  ${chalk.bold(`${logo} now dns rm`)} <id>
  ${chalk.dim('Options:')}
    -h, --help              output usage information
    -c ${chalk.bold.underline('FILE')}, --config=${chalk.bold.underline(
    'FILE'
  )}  config file
    -d, --debug             debug mode [off]
    -t ${chalk.bold.underline('TOKEN')}, --token=${chalk.bold.underline(
    'TOKEN'
  )} login token
  ${chalk.dim('Examples:')}
  ${chalk.gray('–')} List all your DNS records
      ${chalk.cyan('$ now dns ls')}
  ${chalk.gray('–')} Add an A record for a subdomain
      ${chalk.cyan(
        '$ now dns add <YOUR DOMAIN> <SUBDOMAIN NAME> A <RECORD VALUE>'
      )}
      ${chalk.cyan('$ now dns add zeit.rocks api A 198.51.100.100')}
  ${chalk.gray('–')} Add an MX record (@ as a name refers to the domain)
      ${chalk.cyan(
        '$ now dns add <YOUR DOMAIN> @ MX <RECORD VALUE> <PRIORITY>'
      )}
      ${chalk.cyan('$ now dns add zeit.rocks @ MX mail.zeit.rocks 10')}
`)

// ... to this.

console.log(chalk`
  {bold ${logo} now dns ls} [domain]
  {bold ${logo} now dns add} <domain> <name> <A | AAAA | ALIAS | CNAME | TXT> <value>
  {bold ${logo} now dns add} <domain> <name> MX <value> <mx_priority>
  {bold ${logo} now dns add} <domain> <name> SRV <priority> <weight> <port> <target>
  {bold ${logo} now dns rm} <id>

  {dim Options:}

    -h, --help              output usage information
    -c {bold.underline FILE}, --config={bold.underline FILE}  config file
    -d, --debug             debug mode [off]
    -t {bold.underline TOKEN}, --token={bold.underline TOKEN} login token

  {dim Examples:}

  {gray –} List all your DNS records

      {cyan $ now dns ls}

  {gray –} Add an A record for a subdomain

      {cyan $ now dns add <YOUR DOMAIN> <SUBDOMAIN NAME> A <RECORD VALUE>}
      {cyan $ now dns add zeit.rocks api A 198.51.100.100}

  {gray –} Add an MX record (@ as a name refers to the domain)

      {cyan $ now dns add <YOUR DOMAIN> @ MX <RECORD VALUE> <PRIORITY>}
      {cyan $ now dns add zeit.rocks @ MX mail.zeit.rocks 10}
`)
```

Qix-:
> Can you cover those last template.js lines?

- `19` would never occur since we bound `start` to 0 or above in the only call in the entire file
- `111`, `140` - :+1:

sindresorhus:
> Further, it allows us to do the escaping of the interpolated values (see `index.js` in `chalkTag`) and the function _always_ has valid input when used as a tag.

That's a good point.

index.js
```diff
@@ -11,19 +13,44 @@ const levelMapping = ['ansi', 'ansi', 'ansi256', 'ansi16m'];
...
+	if (!this || !(this instanceof Chalk) || this.template) {
+		const chalk = {};
+		applyOptions(chalk, options);
+
+		chalk.template = function () {
+			const args = [].slice.call(arguments);
+			return chalkTag.apply(null, [chalk.template].concat(args));
+		};
+
+		Object.setPrototypeOf(chalk, Chalk.prototype);
```
SimenB: Would it be possible to not use `Object.setPrototypeOf` here? Since it's completely unpolyfillable in IE10, it breaks for jest-matchers when trying to use that in the browser.

I know "supporting" older browsers suck, but this is about not failing spectacularily.

Just warpping it in a `if (Object.setPrototypeOf)` might work around the crashing, and then work out all right?

Ref facebook/jest#4074

sindresorhus: Sorry, we're not going to change anything in Chalk because of browsers.

Qix-:
Chalk is not, nor will it ever be, for the browser. Please stop using ANSI escapes in the browser. The browser will never support ANSI escapes.  ANSI escapes are already mis-used on a wide basis. The browser is at the top of the "misused list". The browser console will never support escapes. Use HTML instead.

Thanks <3

SimenB:
@Qix- We don't want to use chalk in the browser. However, we have some code which we want to use in both browser and on node. Browser version currently throws on IE10 and older just doing `require('chalk')`, not trying to actually use it.

We'll just have to work around it though 🙂

Qix-:
Sounds like you should use optional dependencies and `require.resolve()` to check if a dependency exists.

SimenB:
That won't help as it installs just fine, it just explodes runtime.

We're likely to just stub out chalk in the browser bundle, should be straight forward enough (at least for Jest's use case)

SimenB:
Just to follow up in case somebody stumbles across this later: we ended up creating a stub of the chalk API which just returns the string it got as input, and replace the real chalk with it when bundling for the browser. Quite hacky, but not a lot of code.

https://github.com/facebook/jest/commit/51e0a2494233c1e132e5e016646e941ca7afd647

Seems to work, at least :smile:

PR #169, opened 2017-07-01
Add TypeScript type definition

### LINKED ISSUE #148: TypeScript type definition
It would be good to have the typings for chalk distributed with the npm package:

- Easier to use with TypeScript (no [external dependencies](https://www.npmjs.com/package/@types/chalk) needed)
- Easier to ensure that the typings are in sync with the library
- Easier for users to be more confident that the typings are in sync

t-sauer:
Closes #148 

I don't know if there is a better solution but this seems to work pretty good for me.


t-sauer:
Found some small issues. Needs some more work.

t-sauer:
I can't get the `chalk.constructor` to work both with and without `new`. The version right now only works without it.

Qix-:
@t-sauer All of our tests pass. What error are you seeing?

t-sauer:
I put the tests into a Typescript file to see if everything is typed correctly (I assume most cases are covered in there).  
So in the tests sometimes `chalk.constructor` is used with `new` and sometimes without `new`. Right now I get the following error for when it is used with `new`:

```
[ts] Only a void function can be called with the 'new' keyword.
[ts] 'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.
```

I am able to eliminate this error but then the cases where it is used without `new` aren't covered and I get an error there.

Qix-:
Not sure if the constructor will be supported by typescript, then. It's edge-casey anyway, so I would leave it untyped.

sindresorhus:
Can you add a `ts-test.js` file with some chalk usage and have it run in the npm test run script, like https://github.com/avajs/ava/blob/5ae434a07e9fc25a44f3d78d65b6bc949abeb4a4/package.json#L45 So we would know if it were to break in the future.

t-sauer:
I colocated the typings with the test. That's how it is on DT and makes the most sense to me. If you disagree I can change it.

package.json
```diff
@@ -51,8 +51,10 @@
 		"mocha": "*",
 		"nyc": "^11.0.2",
 		"resolve-from": "^3.0.0",
+		"typescript": "^2.4.1",
 		"xo": "*"
 	},
+	"types": "./index.d.ts",
```
sindresorhus: This needs to be updated now

types/index.d.ts
```diff
@@ -0,0 +1,80 @@
...
+export = chalk;
+
+declare function chalk(...text: (string | number)[]): string;
+declare function chalk(text: TemplateStringsArray, ...placeholders: (string | number)[]): string;
+
+declare namespace chalk {
+
+	interface ChalkConstructorOptions {
+		enabled?: boolean;
+		level?: number;
```
sindresorhus: Could we use a TS enum here?

types/index.d.ts
```diff
@@ -0,0 +1,80 @@
+// Type definitions for Chalk 2
+// Definitions by: Thomas Sauer <https://github.com/t-sauer>
+
+export = chalk;
+
+declare function chalk(...text: (string | number)[]): string;
+declare function chalk(text: TemplateStringsArray, ...placeholders: (string | number)[]): string;
```
sindresorhus: I know we allow any value here, but I think we should enforce a string as TS users want strict types. On line 6 and 7.

types/index.d.ts
```diff
@@ -0,0 +1,80 @@
...
+	export const black: typeof chalk;
+	export const red: typeof chalk;
+	export const green: typeof chalk;
+	export const yellow: typeof chalk;
+	export const blue: typeof chalk;
+	export const magenta: typeof chalk;
+	export const cyan: typeof chalk;
+	export const white: typeof chalk;
+	export const gray: typeof chalk;
+	export const grey: typeof chalk;
```
sindresorhus: Can we make `grey` (with `e`) a type alias or some kind of alias of `gray`?

types/index.d.ts
```diff
@@ -0,0 +1,80 @@
...
+	interface ChalkConstructorOptions {
+		enabled?: boolean;
+		level?: number;
+	}
+
+	export function constructor(options?: ChalkConstructorOptions): typeof chalk;
+
+	export let enabled: boolean;
+	export let level: number;
+	export const supportsColor: boolean;
```
sindresorhus: Is it possible to define this as just an re-export of the [`supports-color`](https://github.com/chalk/supports-color) package? I guess this could be useful if users wants to Cmd+click to view source definition or something.

t-sauer: Go to definition always brings you to the type definition file and not the actual implementation so I don't know how helpful it is.
But it would be possible to create a definition for `supports-color` and add it as a dependency.

types/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+// Definitions by: Thomas Sauer <https://github.com/t-sauer>
+
+export = chalk;
+
+declare function chalk(...text: string[]): string;
+declare function chalk(text: TemplateStringsArray, ...placeholders: string[]): string;
+
+declare namespace chalk {
+
+	export enum ChalkLevel {
```
sindresorhus: `ChalkLevel` => `Level`

types/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+		Disabled = 0,
+		Basic = 1,
+		Extended = 2,
+		TrueColor = 3
+	}
+
+	interface ChalkConstructorOptions {
```
sindresorhus: `ChalkConstructorOptions` => `ConstructorOptions`

types/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export = chalk;
+
+declare function chalk(...text: string[]): string;
+declare function chalk(text: TemplateStringsArray, ...placeholders: string[]): string;
+
+declare namespace chalk {
+
+	export enum ChalkLevel {
+		Disabled = 0,
```
sindresorhus: I would call it `None` as it's not technically disabled, but rather no color support.

types/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+		Basic = 1,
+		Extended = 2,
+		TrueColor = 3
```
sindresorhus: @Qix- Are the enum names ok to you?

types/test.ts
```diff
@@ -0,0 +1,37 @@
+import * as chalk from '..';
+
+chalk.underline('foo');
+chalk.red('foo');
+chalk.bgRed('foo');
+
+const name = 'Josh';
+const number = 0;
```
sindresorhus: Why was this added?

types/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	interface ChalkConstructorOptions {
+		enabled?: boolean;
+		level?: ChalkLevel;
+	}
+
+	export function constructor(options?: ChalkConstructorOptions): typeof chalk;
+
+	export let enabled: boolean;
```
SamVerschueren: Although these are defined as `let`, it's not possible to set the property manually. I tried it in the tests but it failed.

types/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+	export let level: ChalkLevel;
```
SamVerschueren: Same as `enabled`.

types/test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+chalk.constructor({level: 1});
+const ctx = chalk.constructor({level: chalk.ChalkLevel.TrueColor });
+ctx('foo');
+ctx.red('foo');
+ctx`foo`;
+
+chalk.level === chalk.ChalkLevel.Extended;
+
+chalk.enabled;
```
SamVerschueren: Test setting this one (and `level` as well)

`chalk.enabled = true`

types/index.d.ts
t-sauer: chalk is imported as an ES6 module where this is disallowed. If it is imported via `import chalk = require('..');` it is possible to set those values. I can change the test if it is wanted.

sindresorhus: @t-sauer The following is valid in the JS code and should be allowed in TS too:

```js
import chalk from 'chalk';
chalk.enabled = false;
chalk.level = 2;
```

t-sauer: The TS guys seem to disagree: https://github.com/Microsoft/TypeScript/issues/6751

I would also agree with your opinion but I don't know how I can get Typescript to like it.

t-sauer: Never mind, I found a solution. Will push it in a couple of minutes.

SamVerschueren:
@sindresorhus Currently it's not possible to import chalk like this `import chalk from 'chalk';` because there is no default export. I think if we add a default export, the definition could be written a lot more cleaner. Same like we do with AVA.

t-sauer:
@SamVerschueren Exactly what I just changed. That also solved the problem of setting level and enabled.

types/index.d.ts
```diff
@@ -0,0 +1,87 @@
+// Type definitions for Chalk
+// Definitions by: Thomas Sauer <https://github.com/t-sauer>
+
+export default chalk;
```
SamVerschueren: This is currently incorrect because we aren't yet exporting the `default` property. If we agree upon doing that (not my call to make), I'd suggest switching to an interface.

```
export default interface Chalk {
    (...text: string[]): string;

    enabled: boolean;
    level: Level;
    readonly supportsColor: boolean;

    hex(color: string): Chalk;
    // ...
}
```

Gives much better results in VS Code and looks way cleaner as well.

sindresorhus: @t-sauer ⬆️ What do you think?

t-sauer: Looks way better than my current solution. Will continue on it later 👍

j-f1:
What’s the status on this?

sindresorhus:
@j-f1 https://twitter.com/slicknet/status/782274190451671040

PR #174, opened 2017-07-16
Demonstrate crash when using Truecolor methods and color is unsupported

vadimdemedes:
Failing test for https://github.com/chalk/chalk/issues/175.

### DIFF
--- a/test.js
+++ b/test.js
@@ -2,6 +2,7 @@
 const assert = require('assert');
 const importFresh = require('import-fresh');
 const resolveFrom = require('resolve-from');
+const execa = require('execa');
 
 // Spoof supports-color
 require.cache[resolveFrom(__dirname, 'supports-color')] = {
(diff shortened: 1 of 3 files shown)


test.js
```diff
@@ -202,6 +203,13 @@ describe('chalk.level', () => {
 		assert.equal(chalk.level, 1);
 		chalk.level = oldLevel;
 	});
+
+	it('should disable colors if they are not supported', () => {
+		return execa('node', ['fixture'])
+			.then(result => {
+				assert.equal(result.stdout, 'test');
+			});
+	});
```
Qix-: Does this guarantee that the spawned `node` instance is not supported? I'm failing to see how this is any different than the tests now.

vadimdemedes: Hm, not sure what you mean. It crashes for me every time. Could you explain?

Qix-: So recently we had to mock `supports-color` in our tests since CIs handled them differently. Does this test guarantee that no matter what the environment supports, it runs without colors?

vadimdemedes: Yes, I think so, because it fails everywhere: on CI and on my laptop. It fails when publishing a package with `np`, spawning a process manually in repl, in tests. But if you run `node fixture` directly, it works as expected. I was able to reproduce the crash every time, so I think this failing test is stable.

kevva: It will always fail at https://github.com/chalk/supports-color/blob/master/index.js#L44-L46 when running as a spawned instance because `stdout` in [options.stdio](https://nodejs.org/api/child_process.html#child_process_options_stdio) being `pipe`. `execa('node', ['fixture'], {stdio: ['pipe', 'inherit', 'pipe')` would pass. So yeah, this test is good.

PR #182, opened 2017-07-23
Move testing to `ava`

### LINKED ISSUE #181: Move to AVA for testing
Let's move all testing related things into a `test` directory. The current Mocha test groups should just be separate test files.

kevva:
* Renamed all test titles beginning with `should` or `shouldn't`
* `chalk` => `m`

Looks like three tests that passed before fails now. Not really sure why.

Fixes #181.


test.js
```diff
@@ -1,353 +0,0 @@
...
-		hasBasic: true,
-		has256: true,
-		has16m: true
-	}
-};
-
-const chalk = require('.');
-
-console.log('TERM:', process.env.TERM || '[none]');
-console.log('platform:', process.platform || '[unknown]');
```
kevva: Do we want these still?

kevva:
Ok, looks like the previous tests were false positives because of `assert.equal` not being that strict. `chalk.red()` doesn't return an empty string, but `false`.

sindresorhus:
> `chalk.red()` doesn't return an empty string, but `false`.

Ugh, that's a bug. Chalk should always return a string type.

test/chalk.js
```diff
@@ -0,0 +1,104 @@
...
+
+// Spoof supports-color
+require.cache[resolveFrom(__dirname, 'supports-color')] = {
+	exports: {
+		level: 3,
+		hasBasic: true,
+		has256: true,
+		has16m: true
+	}
+};
```
sindresorhus: Can you move this out into a helper that accepts `__dirname` as an argument?

sindresorhus: Actually, can't we just do `process.env.color = '16m';`? // @Qix-

Qix-: Potentially, but this is the only way to really guarantee what we want. I feel more comfortable with this since `supports-color`'s behavior might change.

sindresorhus:
@kevva Is this ready to :shipit: ?

kevva:
Unless we want to fix the failing test in here? Else, yes.

sindresorhus:
Thanks Kevin

kevva:
FYI, I just think the test is invalid:

screen shot 2017-07-23 at 22 21 12

sindresorhus:
@kevva When you moved from Mocha to AVA, you reduced the indentation and that was signification as the fixture we used was using the test indentation against a fixture. Fixed in https://github.com/chalk/chalk/commit/4c4ba2df3d7630fd16f0c49c4b7c923109338807

kevva:
Ohh, right. Forgot is was inside a `describe` before. Good catch.

PR #183, opened 2017-07-23
Return empty `string` when there are no arguments

kevva:
Fixes https://github.com/chalk/chalk/pull/182#issuecomment-317263716.

### DIFF
--- a/index.js
+++ b/index.js
@@ -152,7 +152,7 @@ function applyStyle() {
 	// Support varags, but simply cast to string in case there's only one arg
 	const args = arguments;
 	const argsLen = args.length;
-	let str = argsLen !== 0 && String(arguments[0]);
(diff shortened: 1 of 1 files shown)


Qix-:
Could just fix this by doing `if (argsLen === 0) return '';`, and skip a bunch of branching below.

index.js
```diff
@@ -154,6 +154,10 @@ function applyStyle() {
 	const argsLen = args.length;
 	let str = argsLen !== 0 && String(arguments[0]);
```
kevva: This check is probably moot now, want me to remove it?

Qix-: Yeah should be fine. `String(arguments[0])` is going to evaluate to `String(undefined)`, which will just evaluate to `"undefined"` - that is, in the case that `argsLen === 0`.

kevva: Done.

PR #191, opened 2017-07-31
Fix template literals

### LINKED ISSUE #175: Truecolor methods crash when color is unsupported
When `supports-color` returns `false`, `supportsColor.level` equals `undefined` here: https://github.com/chalk/chalk/blob/master/index.js#L22. This makes `this.level` equal `undefined` here https://github.com/chalk/chalk/blob/master/index.js#L75 and as a result, crashes when color is unsupported. This can be reproduced by executing a script with chalk output using `child_process.spawn()` and piping output to stdout.

The fix is to change this:

```js
obj.level = options.level === undefined ? supportsColor.level : options.level;
```

to this:

```js
obj.level = options.level === undefined ? (supportsColor.level || 0) : options.level;
```

I stumbled upon this issue when tests were passing in terminal and on Travis CI, but were failing when run by `np`.

I submitted a PR with a failing test here https://github.com/chalk/chalk/pull/174. If I didn't miss anything, I will modify this PR to fix the problem.

Note, that "regular" (e.g. `chalk.blue()`) methods don't fail.

### LINKED ISSUE #176: Got a error when use `chalk.bgHex()` with arg `--no-color`
```js
try {
	chalk.bgHex('#cccccc')('message');
} catch (ex) {
	console.error(ex);
}
```
```bash
E:\work\gulp-reporter (master) (gulp-reporter@2.1.0)
$ node -v
v8.1.2

E:\work\gulp-reporter (master) (gulp-reporter@2.1.0)
$ node test.js --no-color
TypeError: Cannot read property 'hex' of undefined
    at Function.<anonymous> (E:\work\gulp-reporter\node_modules\chalk\index.js:98:57)
    at errors.map (E:\work\gulp-reporter\lib\formater.js:240:19)
    at Array.map (native)
    at formater (E:\work\gulp-reporter\lib\formater.js:235:18)
    at reporter (E:\work\gulp-reporter\lib\reporter.js:20:10)
    at getErrors.then.errors (E:\work\gulp-reporter\lib\gulp-plugin.js:19:19)

```

### LINKED ISSUE #177: Newline character (\n) in tagged template literals becomes "n"
From the comment here: https://github.com/chalk/chalk/issues/12#issuecomment-315685346
> When using tagged template literals, `\n` does not work => `n` is output and there is no line break.

I just ran into this myself, so I figured I'd do as @Qix- asked.

------

It looks like part of your processing is eating the `\`. Multi-line template literals (or anything similar in other languages) always seem like a cool idea until you have code like this...

```js
function run() {
	if (something) {
		if (somethingElse) {
			console.log(`this is the first line
			this is the second`);
		}
	}
}
```

...producing the output...

```
this is the first line
			this is the second
```

...which isn't optimal. You could pull the second line all the way to the margin, but that'd never fly with formatters or linters.

Maybe there's a better way for chalk to detect character escapes (certain character escapes) and not eat them?

As a workaround, it is possible to do something like this, though it's not very attractive:

```js
console.log(chalk`{red NO} but...${'\n'}really, {green YES}`);
```

### LINKED ISSUE #184: The {} placeholder removes the linebreak that's just before it
In a multiline chalk template string, the {} placeholder removes the linebreak that's just before it (i.e. if it is at the beginning of any line).

So for example:
``` 
chalk`The first line
{red The second line in red}`
```
produces:
```
The first lineThe second line in red
```
when it should produce:
```
The first line
The second line in red
```

This bug does not happen if the {} placeholder is located anywhere after the beginning of a line.

I don't know if this is related to #177 or #12 but at first glance this seems like a distinct bug to me.

### LINKED ISSUE #186: Add custom theme support to template literals
It'd be sick if we could do something like

```javascript
const myTheme = require('chalk').constructor(); // no `new`

myTheme.rose = myTheme.hex('#F6D9D9');

const text = myTheme`This is some {rose Rose Text}.`;
```

### LINKED ISSUE #187: Root chalk function doesn't support multiple arguments if used in template mode
When creating a chalk object with template functionality:

```javascript
const ctx = chalk.constructor({level: 0}); // no new; templates enabled
```
calling a style with multiple arguments will concatenate them with spaces

```javascript
const str = ctx.red('hello', 'there'); //-> "hello there"
```

but calling the root function doesn't:

```javascript
const str = ctx('hello', 'there'); // -> "hello"
```

### LINKED ISSUE #194: Interpolated template expressions that evaluate to `undefined` or `null` throw exception
Getting the following error message with template literals:

```
> chalk`hello ${undefined}`

TypeError: Cannot read property 'toString' of undefined
    at chalkTag (/src/bitlang/bootstrap/node_modules/chalk/index.js:210:25)
    at Chalk.chalk.template (/src/bitlang/bootstrap/node_modules/chalk/index.js:36:20)
```

Qix-:
So the old templating code wasn't really build for how we are utilizing template literals today. A lot of the bugs we faced were the result of unforeseen/unspec'd requirements that were imposed upon the existing code.

I took a somewhat different approach - a manual parser is all fine and good, but it's tricky to maintain, especially when lots of things pop up and once or when the scope/intent of the code changes drastically (as can be seen by the bug tracker :taco:).

It's served us well thus far but I felt it was time to give it a little bit of a ✨ makeover ✨.

As a result, it fixes a bunch of bugs and introduces the ability to add custom theme values to your chalk instance and use them in templates (#186).

```javascript

const theme = chalk.constructor();
theme.rose = theme.hex('#F69D9D');

console.log(theme`Here is some {rose rosy text!}`);
```

I feel like I'll need to justify using regular expressions here instead of manual parsing:

1. They're being (ab)used mainly as tokenizers
2. The longest one is just a series of smol patterns or'd together
3. It greatly simplifies the codebase and keeps maintenance costs low (you can drop them into regex101.com and make changes as necessary)

By using templates, you're already [incurring quite a bit of overhead](https://stackoverflow.com/a/29083467/510036) in tight loops, so the template literal parsing code is going to be of negligible overhead.

Don't be fooled by the diff of `templates.js` - it's a complete rewrite. Don't waste your time trying to parse the diff 😅  Not sure why Git isn't better at detecting rewrites.

Closes #186
Fixes #184 
Fixes #187 
Fixes #176 
Fixes #175 
Fixes #177 
Fixes #194

cc @vadimdemedes

---

Note to merger: It's probably best _not_ to squash this particular PR. I took a bit of time to rebase it prior to PRing in order to have a nice and clean fast-forward/rebase. cc @sindresorhus

### DIFF
--- a/index.js
+++ b/index.js
@@ -207,8 +207,8 @@ function chalkTag(chalk, strings) {
 	const parts = [strings.raw[0]];
 
 	for (let i = 1; i < strings.length; i++) {
-		parts.push(args[i - 1].toString().replace(/[{}\\]/g, '\\$&'));
-		parts.push(strings.raw[i]);
+		parts.push(String(args[i - 1]).replace(/[{}\\]/g, '\\$&'));
+		parts.push(String(strings.raw[i]));
 	}
 
 	return template(chalk, parts.join(''));
--- a/index.js
+++ b/index.js
@@ -197,12 +197,13 @@ function applyStyle() {
 }
 
 function chalkTag(chalk, strings) {
-	const args = [].slice.call(arguments, 2);
-
 	if (!Array.isArray(strings)) {
-		return strings.toString();
+		// If chalk() was called by itself or with a string,
+		// return the string itself as a string.
+		return [].slice.call(arguments, 1).join(' ');
 	}
 
+	const args = [].slice.call(arguments, 2);
 	const parts = [strings.raw[0]];
 
 	for (let i = 1; i < strings.length; i++) {
--- a/index.js
+++ b/index.js
@@ -205,7 +205,7 @@ function chalkTag(chalk, strings) {
 	const parts = [strings.raw[0]];
 
 	for (let i = 1; i < strings.length; i++) {
-		parts.push(args[i - 1].toString().replace(/[{}]/g, '\\$&'));
+		parts.push(args[i - 1].toString().replace(/[{}\\]/g, '\\$&'));
 		parts.push(strings.raw[i]);
 	}
 
--- a/index.js
+++ b/index.js
@@ -19,7 +19,8 @@ function applyOptions(obj, options) {
 	options = options || {};
 
 	// Detect level if not set manually
-	obj.level = options.level === undefined ? supportsColor.level : options.level;
+	const scLevel = supportsColor ? supportsColor.level : 0;
+	obj.level = options.level === undefined ? scLevel : options.level;
 	obj.enabled = 'enabled' in options ? options.enabled : obj.level > 0;
 }
 
--- a/templates.js
+++ b/templates.js
@@ -1,162 +1,128 @@
 'use strict';
+const TEMPLATE_REGEX = /(?:\\(u[a-f0-9]{4}|x[a-f0-9]{2}|.))|(?:\{(~)?(\w+(?:\([^)]*\))?(?:\.\w+(?:\([^)]*\))?)*)(?:[ \t]|(?=\r?\n)))|(\})|((?:.|[\r\n\f])+?)/gi;
+const STYLE_REGEX = /(?:^|\.)(\w+)(?:\(([^)]*)\))?/g;
+const STRING_REGEX = /^(['"])((?:\\.|(?!\1)[^\\])*)\1$/;
+const ESCAPE_REGEX = /\\(u[0-9a-f]{4}|x[0-9a-f]{2}|.)|([^\\])/gi;
 
-function data(parent) {
-	return {
-		styles: [],
-		parent,
-		contents: []
-	};
+const ESCAPES = {
+	n: '\n',
+	r: '\r',
+	t: '\t',
+	b: '\b',
+	f: '\f',
+	v: '\v',
+	0: '\0',
+	'\\': '\\',
+	e: '\u001b',
+	a: '\u0007'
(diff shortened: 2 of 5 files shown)


sindresorhus:
Btw, `Fixes #184, #187, #176, #175, #177, and #194.` won't work. You need `Fixes` in front of each issue number. Same with the commit titles. Can you also uppercase the first character in commit titles? That's the style I'm using now.

bokub:
Hi, do you plan to merge/publish this soon?
The bugs fixed by this PR are quite annoying...

Thanks :+1:

Qix-:
@bokub commenting on PRs asking when they'll be fixed only 3 days after they've been submitted doesn't help get them merged.

We know these bugs are annoying - we use chalk too. Keep in mind we do this in our free time.

Qix-:
@sindresorhus could have sworn Github detected multiple fixes in commits historically.

sindresorhus:
> Auto-close doesn't add anything other than convenience.

Yes, they do:

screen shot 2017-08-05 at 23 36 39

Qix-:
Published as `2.1.0`. Migration to `Map` will happen later - we just needed these pushed out pronto.

Minor release due to templating rewrite, change to template handler function (which we've already seen is being extended), and the ability to add custom themed styles to a Chalk context and have them appear in template literals.

Should be trivial to update since Chalk 2 is so new.

PR #193, opened 2017-08-01
Add related packages in the readme

bokub:
- Added [chalk-animation](https://github.com/bokub/chalk-animation) as suggested by @sindresorhus [here](https://github.com/sindresorhus/module-requests/issues/95#issuecomment-319166247)
- Added [gradient-string](https://github.com/bokub/gradient-string) because it also uses chalk but provides features unavailable in chalk (gradients)


sindresorhus:
You should submit them to https://github.com/sindresorhus/awesome-nodejs too. Note that there should be one package per pull request when submitting there. --> https://github.com/sindresorhus/awesome-nodejs/blob/master/contributing.md

PR #205, opened 2017-09-01
Paraquire - paranoidal require

nickkolok:
NPM ecosystem is [quite unresistant](https://github.com/ChALkeR/notes/blob/master/Gathering-weak-npm-credentials.md) against dependency attacks. It means that if a hacker gets control over one small abandoned package, the hacker gets control over all it's dependents, dependents of dependents, etc.
So, hijacking only one of small `chalk`'s dependencies will be very crucial for whole NPM ecosystem. This PR improves `chalks`'s security by jailing most dependencies into `paraquire` - secure [alternative](https://github.com/nickkolok/paraquire) for standard `require`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,10 @@
 'use strict';
-const escapeStringRegexp = require('escape-string-regexp');
-const ansiStyles = require('ansi-styles');
-const supportsColor = require('supports-color');
+
+const paraquire = require('paraquire')(module);
+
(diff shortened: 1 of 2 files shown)


Qix-:
:-1: Have never seen this happen, and we control all of our dependencies (e.g. @sindresorhus and I own all of them).

Further, this would introduce yet another dependency that we do not control, and at 32+ million downloads a month I don't even want to begin chancing something like this.

Sorry but I have to decline.

Issue #215, opened 2017-10-18
Minor version 2.2 is breaking for TypeScript consumers


felixfbecker:
https://github.com/sourcegraph/javascript-typescript-langserver/issues/371
https://travis-ci.org/sourcegraph/javascript-typescript-langserver/jobs/289329597#L1487-L1490

```
src/logging.ts(72,40): error TS2339: Property 'grey' does not exist on type 'typeof "/home/travis/build/sourcegraph/javascript-typescript-langserver/node_modules/chalk/types/...'.
src/logging.ts(80,40): error TS2339: Property 'bgCyan' does not exist on type 'typeof "/home/travis/build/sourcegraph/javascript-typescript-langserver/node_modules/chalk/types/...'.
src/logging.ts(88,40): error TS2339: Property 'bgYellow' does not exist on type 'typeof "/home/travis/build/sourcegraph/javascript-typescript-langserver/node_modules/chalk/types/...'.
src/logging.ts(96,40): error TS2339: Property 'bgRed' does not exist on type 'typeof "/home/travis/build/sourcegraph/javascript-typescript-langserver/node_modules/chalk/types/...'.
```

Before 2.2, with `@types/chalk`, it was possible to import chalk like this:

https://github.com/sourcegraph/javascript-typescript-langserver/blob/6fcc4e1df9283dbd84db9411aa26e9f49556cb50/src/logging.ts#L2

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/05ac46f9a9307d43e082b7cec4b32f40df3a3b58/types/chalk/index.d.ts#L6

2.2 brings its own types, which override `@types/chalk`, but they don't allow this import (only a defaut import):

https://github.com/chalk/chalk/blob/d86db88e778fa856f4d6f5f68c588750ca06b822/types/index.d.ts#L90

Although it seems to work at runtime and appears semantically correct / correct in regard to the JS:

https://github.com/chalk/chalk/blob/d86db88e778fa856f4d6f5f68c588750ca06b822/index.js#L219

sindresorhus:
Sorry about the trouble, but this seems more like a problem with the DefinitelyTyped way of handling typings. Projects can't really take into account unofficial type definitions. We went with a default export now as that's more future proof. It's not a breaking change, as Chalk didn't officially come with typings before.

Maybe you should open an issue on TypeScript and request that `@types/chalk` kind of typings takes priority over ones that come with packages?

sindresorhus:
Obligatory:

[...]

PR #216, opened 2017-10-18
Remove arbitrary value expression in the TS type definition

calebboyd:
based on upcoming breaking change in typescript 2.6 https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#arbitrary-expressions-are-forbidden-in-export-assignments-in-ambient-contexts

Missed this! 😬

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -86,5 +86,6 @@ export interface Chalk {
 	bgWhiteBright: Chalk;
 }
 
-declare function chalk (): any;
-export default chalk as Chalk;
+declare const chalk: Chalk
+
+export default chalk


calebboyd:
re #215 Should we also add each property as a named export?

types/index.d.ts
```diff
@@ -85,6 +85,5 @@ export interface Chalk {
 	bgCyanBright: Chalk;
 	bgWhiteBright: Chalk;
 }
-
-declare function chalk (): any;
-export default chalk as Chalk;
+declare const chalk: Chalk
```
Qix-: New line before the `declare` statement, please :)

calebboyd: I've updated both PRs.

calebboyd:
Closed in favor of #217

PR #217, opened 2017-10-18
TypeScript fixes

calebboyd:
https://github.com/chalk/chalk/pull/217#issuecomment-337795227


Qix-:
@calebboyd Does this PR address https://github.com/chalk/chalk/issues/215#issuecomment-337601399?

calebboyd:
It would solve the need to find and replace the import method. (it adds named exports)

But as for the example:
```typescript
function logBold(msg): string {
  return chalk.bold(msg)
}
```
`logBold` accepts an implicit any, which isn't supported by any signature that we have defined...
If we wanted to support that we could add the type:
`(...text: any[]): string` here:
https://github.com/chalk/chalk/blob/d86db88e778fa856f4d6f5f68c588750ca06b822/types/index.d.ts#L19

AFAIK the only supported behavior is accepting string rest args.

Qix-:
@calebboyd is there a way to specify `string | string[]`?

calebboyd:
Since its rest args it has to be an array type.

eg. the following will work.

```typescript
chalk.bold('a')
chalk.bold('a','b','c')
```

calebboyd:
@Qix- I was able to modify the overload order to provide better implicit results.

sindresorhus:
I've changed my mind (Sorry @calebboyd ...). I don't think we should do the named exports. I thought it could be a nice transition step, but it's a day later now and it doesn't seem to be a big problem, and the few that had a problem have fixed it and moved on, so I think it's better for everyone that we just stick to a default export than inconvenience everyone with two ways of using Chalk, one of which is IMHO a bad practice, as I've elaborated on in https://github.com/chalk/chalk/issues/215#issuecomment-337459457.

calebboyd:
Okay Cool. (I totally agree) I'll Clean it up. The other fixes are still valid I think

types/index.d.ts
```diff
@@ -13,20 +13,26 @@ export interface ChalkOptions {
...
+	has256: boolean;
+	has16m: boolean;
+}
+
+export interface Chalk {
 	(...text: string[]): string;
 	(text: TemplateStringsArray, ...placeholders: string[]): string;
-	constructor: Chalk;
+	new (options?: ChalkOptions): Chalk;
+	(options: ChalkOptions): Chalk;
```
sindresorhus: Why are these added here? They're only valid for `chalk.constructor()`

types/index.d.ts
```diff
@@ -86,5 +92,6 @@ export interface Chalk {
 	bgWhiteBright: Chalk;
 }
 
-declare function chalk (): any;
-export default chalk as Chalk;
+declare const chalk: Chalk & { supportsColor: ColorSupport };
```
sindresorhus: Why was `& { supportsColor: ColorSupport }` moved out of the `Chalk` interface? Or rather, what's the benefit of having it here?

types/index.d.ts
```diff
@@ -13,20 +13,26 @@ export interface ChalkOptions {
 	level?: Level;
 }
 
-export interface Chalk {
+export interface ChalkConstructor extends Chalk {
```
alan-agius4: Why not use classes rather than Interfaces when the Object is clearly a class here?

alan-agius4: This doesn't look right, Is `chalk` a class or an object callable with `new`?

types/index.d.ts
calebboyd: I missed moving them when I pulled ChalkConstructor out 😬

types/index.d.ts
calebboyd: AFAICT `supportsColor` only exists on the default export? If it is left in the `Chalk` interface it will show up all over where it doesn't exist.

SamVerschueren: You mean that this would be valid as well?

```js
chalk.red.supportsColor
```

I think you're right, it shouldn't.

types/index.d.ts
```diff
@@ -13,20 +13,24 @@ export interface ChalkOptions {
...
 	enabled: boolean;
 	level: Level;
-	supportsColor: {
-		level: Level;
-		hasBasic: boolean;
-		has256: boolean;
-		has16m: boolean;
-	};
 	rgb(r: number, g: number, b: number): Chalk;
 	hsl(h: number, s: number, l: number): Chalk;
```
SamVerschueren: Instead of defining that all these methods return `Chalk`, you can also say that they return `this`.

```js
rgb(r: number, g: number, b: number): this;
```

That's also what TS generates for the type definitions. Just another approach though.

types/index.d.ts
sindresorhus: That's a good point and I agree, but shouldn't that apply to `constructor`, `enabled`, and `level` too then?

SamVerschueren: I agree, but that's not what it looks like at runtime.

These things are possible:
`chalk.blue.enabled`
`chalk.blue.constructor`
`chalk.blue.level`

So if it shouldn't be possible, we will have to rewrite some internals so that the type definitions match what we have at runtime.

calebboyd: @SamVerschueren 
>You mean that this would be valid as well?
 Yeah

or it could be changed to behave like `enabled` and `level`

types/index.d.ts
calebboyd: Updated

types/index.d.ts
sindresorhus: Hmm, there might be some use-case where people want to do:

```js
export const color = chalk.blue.bold;
```

```js
import {color} from '.';

color.enabled = false;
```

calebboyd: @sindresorhus, good use-case -- and that is supported with these typings. But `color.supportsColor` would not be (nor does it exist at runtime).

should `supportsColor` be added to the chained/built chalk properties?

dwieeb:
Chiming in here:

Each "color" is an object with a getter: https://github.com/chalk/chalk/blob/f653b061d6fbdb1c7224f7d80476391202c47877/index.js#L55-L64

Setting these values results in a run-time error: `TypeError: Cannot set property green of #<Chalk> which has only a getter`.

So these typings should change from:

```ts
	reset: this;
	bold: this;
	dim: this;
...
```

to:

```ts
	readonly reset: this;
	readonly bold: this;
	readonly dim: this;
...
```

calebboyd:
@dwieeb You're right!

types/index.d.ts
sindresorhus: Only if we add it in the runtime, which is a separate discussion to this PR. I'm undecided whether it makes sense though.

sindresorhus:
@calebboyd Can you fix the merge conflict from https://github.com/chalk/chalk/commit/dc092b4a5f5ca77dd1e22607cdf2fdd388803064 ?

sindresorhus:
Thank you so much for helping out with this @calebboyd :)

PR #220, opened 2017-10-21
Fix export typing

pelotom:
In TypeScript 2.6 the prior typing fails to compile with

```
export default chalk as Chalk;
               ^^^^^^^^^^^^^^
The expression of an export assignment must be an identifier or qualified name in an ambient context.
```


pelotom:
Duplicate of #217. Oops.

PR #221, opened 2017-10-21
fix(typings): fix error,  the expression of an export assignment must…

alan-agius4:
… be an identifier orqualified name in an ambient context when using TypeScript `2.6.0-rc`

More information https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#arbitrary-expressions-are-forbidden-in-export-assignments-in-ambient-contexts

Closes: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20777

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -86,5 +86,5 @@ export interface Chalk {
 	bgWhiteBright: Chalk;
 }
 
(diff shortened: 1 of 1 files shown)


pelotom:
This is superseded by #217.

PR #222, opened 2017-10-24
Add .visible 

### LINKED ISSUE #192: Add .visible()
Just ran into the use case where it'd be quite useful to have something like `chalk.visible(...)`, where the contents are only emitted if chalk is enabled and has a non-zero level.

This is particularly for borders and whatnot that are purely cosmetic and thus don't need to even appear if chalk is not outputting things.

In the event chalk is disabled, a call to `visible()` returns an empty string.

Thoughts?

danthegoodman:
Add `.visible` for emitting text only when enabled (fixes #192)

This resulted in another property that is tracked across builder instances. 

- [X] Documentation updated
- [X] Tests updated
- [X] Typescript definition updated
- [X] Typescript definition tests updated

### DIFF
--- a/index.js
+++ b/index.js
@@ -63,6 +63,13 @@ for (const key of Object.keys(ansiStyles)) {
 	};
 }
 
+styles.visible = {
+	get() {
+		this._emptyIfNotVisible = true;
+		return build.call(this, this._styles ? this._styles : [], 'visible');
+	}
+};
+
 ansiStyles.color.closeRe = new RegExp(escapeStringRegexp(ansiStyles.color.close), 'g');
 for (const model of Object.keys(ansiStyles.color.ansi)) {
 	if (skipModels.has(model)) {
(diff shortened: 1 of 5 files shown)


Qix-:
Beautiful. Thank you so much :)

Qix-:
Bit of a bug with it. Here's a repro (I'll fix it):

```
chalk.red('hello'); //-> "\x1b[31mhello\x1b[0m"
chalk.enabled = false;
chalk.red('hello'); //-> "hello"
chalk.visible('hello'); //-> ""
chalk.red('hello'); //-> ""     :(
```

Qix-:
Fix/test landed in e2a4aa427568ff1c5d649739c4d1f8319cf0d072. Thanks again! :)

danthegoodman:
@Qix- Oh dang. Good catch! Thanks for fixing that! 😄

PR #229, opened 2017-11-21
Fix windows 10 bold sequence

Vap0r1ze:
This is a problem I've had with a good handful of my projects. After some super deep investigation ~~/s~~, I discovered Windows 10 doesn't support the standard bold close sequence like everything else.

I added a Windows 10 check to set the closing sequence to a full reset, in the same manner the blue is corrected to bright blue.

I console logged these 3 tagged literals in both versions:
- `[{red Good morning, {bold Vap0r1ze}.}]`
- `Please {bold run} {inverse npm i {bold chalk}}`
- `__{underline You have {bold 22} credits}__`

image
*Top is the new version, bottom is the old*

This is how the it looks on Windows 7 & 8 (old and new)
image

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,10 +2,12 @@
 const escapeStringRegexp = require('escape-string-regexp');
 const ansiStyles = require('ansi-styles');
 const supportsColor = require('supports-color');
+const os = require('os')
 
 const template = require('./templates.js');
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
This has already been fixed in the latest Windows 10 update: https://github.com/Microsoft/WSL/issues/2174

Issue #31, opened 2014-07-07
Possible error created in linux


AndrewRayCode:
We are seeing this happen in `chalk`:

```
node_modules/gulp-ruby-sass/node_modules/chalk/node_modules/escape-string-regexp/index.js:4
throw new TypeError('Expected a string');
    ^
```

This was not previously happening. Appears to be related to the changes made 3 days ago. I cannot reproduce this in OSX

The command we're running is gulp sass, which apparently depends on chalk.

netpoetica:
FWIW my team and I are seeing the exact same error, seems to be something to do with npm running test command, running mocha w/ ts-node runtime TypeScript compilation.

```
➜  node -v
v8.6.0
➜  npm -v
5.4.2
```

Will update here if we solve it. But this one is pretty nasty!

PR #231, opened 2017-11-30
escape-string-regexp throws - make sure value passed in str before ca…

netpoetica:
This function should be used responsibly, because escape-string-regexp throws if the value is not a string. Because this module is used so deeply, this can lead to some very cryptic errors from this uncaught exception, that will leave most developers digging 3,4, 5 modules deep to figure out what's causing it.

See:

https://github.com/chalk/chalk/issues/31
https://github.com/TypeStrong/ts-node/issues/468#issuecomment-348184701

It doesn't fix the root cause of the issue, which seems to be conflicting versions of chalk being loaded at the same time in ts-node+mocha tests, but it is more responsible methinks

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,7 +53,10 @@ if (isSimpleWindowsTerm) {
 }
 
 for (const key of Object.keys(ansiStyles)) {
-	ansiStyles[key].closeRe = new RegExp(escapeStringRegexp(ansiStyles[key].close), 'g');
+	// Module escape-string-regexp throws if value entered is not a string.
+	if (typeof ansiStyles[key].close === 'string') {
+		ansiStyles[key].closeRe = new RegExp(escapeStringRegexp(ansiStyles[key].close), 'g');
+	}
 
 	styles[key] = {
 		get() {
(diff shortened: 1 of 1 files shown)


sindresorhus:
This is definitely not the correct solution. The `.close` methods are guaranteed to be there, so it makes no sense to protect against them not being there. That's just hiding the actual issue.

This seems more like npm resolving the wrong sub-dependency though. Or something else outside of Chalk's control.

The first step would be to submit a failing test here.

netpoetica:
I understand. At runtime, you really don't know for sure what will or wont be there though, so it makes sense to check it first before calling a function that you know can throw. Maybe there should be a default value if string is undefined in those cases?

This issue arose because `elasticsearch` module uses 1.1.3 and `ts-node` uses 2.3.0 - at transpilation time in ts-node, the values returned by ansi-styles are as though they are to look for chalk 2.3.0, but then they get loaded into 1.1.3 and this causes the string passed to escape-string-regexp to be undefined. I imagine there is some issue with ts-node that causes it to cache require of ansi-styles and then use it in the other version of chalk, but that's just speculation.

Qix-:
> At runtime, you really don't know for sure what will or wont be there though, so it makes sense to check it first before calling a function that you know can throw. 

But... we do. Because we defined them in the module and we control that module.

netpoetica:
Understandable. I tried to recreate this in some way in order to illustrate the issue in an isolated matter, but a couple hours worth of time, and I can't create an isolated test case - sorry!

I tried copying over the exact package.json (for deps and devDeps), creating same dir structure w/ src and test directories, and loading only the elasticsearch function `new Client()` that was the trigger in our other proprietary project, and making a single src and test file w/ the exact same package.json test command, but the effect did not occur again.

I can guarantee that, for some reason, while running unit tests w/ mocha and ts-node in our project, an undefined value is being passed in to that function, causing it to throw and stop tests. As to "why", I'm not sure. If I can manage to get an isolated case together, I will submit a new issue.

In the interim, this one's got me beat!

Qix-:
@netpoetica make sure it's not your package-lock.json or yarn.lock file(s). I have grown to be very untrusting of them as they cause **tons** of problems.

Issue #234, opened 2017-12-13
chalk.enabled=true or { enabled: true} does not work when color support is initially detected as false


nazrhyn:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/chalk/chalk/issues/234/badge.svg" />](https://issuehunt.io/r/chalk/chalk/issues/234)

## Environment
**Chalk:** 2.3.0
**Node.js:** 6.12.0 and 8.9.3

## Issue
**chalk.js**   
```js
const chalk = require('chalk');
chalk.enabled = true;
console.log(chalk`{green hello}`);
```
--or--
```js
const chalk = new require('chalk').constructor({ enabled: true });
```

If you run **chalk.js** as:

```
node chalk.js
```

The output is a green `hello`.

If you run this in a clean environment (which causes **supports-color** to report that colors are not supported):

```
env -i node chalk.js
```

The output is a un-colored `hello`.

## Details
I traced this to https://github.com/chalk/chalk/blob/master/index.js#L177, where `this.enabled` is `false`, and `this` is `builder`. It seems like, somehow, the initial value for `enabled` is not getting changed or is not filtering down to the `enabled` property defined on `builder`.

As expected, using the environment variable `FORCE_COLOR` or the `--color` parameter does work, as it gets `enabled` to the right value, initially.

Interestingly, running it as (to force it on from the start):

```
env -i node chalk.js --color
```
...and then doing:
```js
chalk.enabled = false;
```
..._does_ actually turn it off. Now I'm thinking there's some issue with truthy and falsy. I'm not sure.

## Aside
It would definitely be easier for me to dig more on this if there weren't _so much_ indirection in the way this is implemented :innocent:; but, I'm sure some or most of that is necessary to get this to work so cleanly.

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/chalk/chalk/issues/234).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/chalk/chalk/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

sindresorhus:
Thanks for opening this well-written ticket. Even if you're not able to submit a fix, it would be helpful if you could submit a [failing test](https://github.com/avajs/ava#failing-tests) to make it easier for us or others to look into it.

nazrhyn:
@sindresorhus To make sure I understand what you're asking for:

[...]

PR #235, opened 2017-12-15
Add failing test for #234

nazrhyn:
## Purpose
Provide failing test for #234.

## Notes
1. **resolve-from** is a little funny. Something like https://github.com/boblauer/mock-require might be a bit simpler/clearer. (Though, I understand you're using your own stuff.)
1. Your test runner doesn't seem to have a construct for grouping tests other than files. In Mocha, if I wanted to run something to build up a single test or set of tests, I'd use a construct like this:

     ```js
     describe('outer grouping', function () {
     
         it('some test', function () {});
     
         describe('inner grouping', function () {
     
             before('build-up before tests in this group', function () {
                 // stuff
             });
     
             it('group test', function () {});
     
             after('tear-down after tests in this group', function () {
                 // stuff
             });
     
         });
     
     });
     ```
   The nested `describe` creates another "group" inside which the `before` / `beforeEach` / `after` / `afterEach` run.   
      
   In this case, I'd like to have a test or a few tests where the mocked return from **supports-color** is different. Not sure how to do that with AVA.

## Actions
What I did was create another file for tests where color is not supported. If you guys understand AVA better, maybe there's a nicer solution for this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
-.idea/
-node_modules/
-coverage/
+node_modules
 yarn.lock
+coverage
 .nyc_output
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-node_modules
+.idea/
+node_modules/
+coverage/
 yarn.lock
-coverage
 .nyc_output
--- a/test/_supports-color.js
+++ b/test/_supports-color.js
@@ -1,15 +1,15 @@
 'use strict';
 const resolveFrom = require('resolve-from');
 
-module.exports = dir => {
-	require.cache[resolveFrom(dir, 'supports-color')] = {
-		exports: {
-			stdout: {
-				level: 3,
-				hasBasic: true,
-				has256: true,
-				has16m: true
-			}
-		}
-	};
+const DEFAULT = {
+	stdout: {
+		level: 3,
+		hasBasic: true,
+		has256: true,
+		has16m: true
+	}
+};
+
+module.exports = (dir, override) => {
+	require.cache[resolveFrom(dir, 'supports-color')] = {exports: override || DEFAULT};
 };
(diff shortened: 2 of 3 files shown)


justsml:
@sindresorhus is behind AVA ... and about 100+ other libraries your using.

Nesting `describe` & `it`'s is a bug, AVA solves this for you.

😇

nazrhyn:
@justsml 

1. Nesting suites (the general term for what calling `describe` creates when using the [BDD interface](https://mochajs.org/#bdd)) and scoping tests and build-up and tear-down to them is a core Mocha feature. Check out the homepage: https://mochajs.org.
1. From my notes:   
      > Not sure how to do that with AVA.   
         
      > If you guys understand AVA better, maybe there's a nicer solution for this.   
      
   I did a little bit more looking. From this issue https://github.com/avajs/ava/issues/222, it looks like test grouping was intentionally left out, so I'm not sure what it is you were going to suggest.

I'm going to guess that you found this page because Google indexed it. :wink:

nazrhyn:
Updated to current master. Not sure if this was here before, but this is happening now:

```
npm WARN deprecated electron@0.4.1: The original electron project has been moved. Visit github.com/logicalparadox/electron for more details.
```

**electron** is a dependency of **matcha**:

```
├─┬ matcha@0.7.0
│ ├─┬ electron@0.4.1
│ │ └─┬ drip@1.1.0
│ │   └── tea-concat@0.1.0
│ └── v8-argv@0.1.0
```

nazrhyn:
@sindresorhus @Qix-   
Hello again; would one of you guys like to merge this so I can stop thinking about it? :smile:

sindresorhus:
Thanks for your patience, @nazrhyn. I totally missed this one.

nazrhyn:
No problem, fellas. I left the tab open at work so I could track it.

PR #238, opened 2017-12-25
Suggested String Template Example

justsml:
Here's a pattern I've been using lately.
Let me know if you think it might be useful to others. 👍

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -83,6 +83,10 @@ RAM: {green ${ram.used / ram.total * 100}%}
 DISK: {rgb(255,131,0) ${disk.used / disk.total * 100}%}
 `);
 
+// ES2015 single-color tagged template literal
+const {yellow, red} = require('chalk');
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks, but we don't consider that a good pattern as the color names are not clear when taken out of context. Suddenly you have a function called `red` deep inside some code path without realizing it's connected to Chalk. This will also not work with ES2015 modules as Chalk is a default export and the colors are not exported individually.

Relevant comment: https://github.com/chalk/chalk/issues/215#issuecomment-337459457

justsml:
Why can't the color functions be exported individually? You don't need to use them, and you can alias them as needed in the `import`. 

Sure, color names may be confusing or out of context in a game, or css-in-js module. 
But for 90% of the work I do, it's definitely obvious `red()` is connected to chalk when it's on every line like:

```js
log(red`FAIL :facepalm:`)
```

Happy Xmas Sindre! 

Thanks for all you do!

PR #246, opened 2018-01-10
Fix IntelliSense

u3u:


sindresorhus:
We already have `export default chalk`. You should probably rather open an issue on InstelliSense.

u3u:
@sindresorhus In vscode, you need to write so that IntelliSense hints correctly: 
```js
const chalk = require('chalk').default
```
image

In fact, it just needs to be written like this: 
```js
const chalk = require ('chalk')
```
image

Define the module export to allow IntelliSense to work correctly

sindresorhus:
The TypeScript definition is intended for TypeScript usage. I don't think we should have to add workarounds to make it work with Intellisense. That's up to VSCode to support.

It might look easy to just add this export, but are you going to do it with every CommonJS module that has TypeScript typings? It usually better to fix the problem at the source. One time. Instead of trying to "fix" every module.

u3u:
@sindresorhus 
But the [official website's template](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-plugin-d-ts.html) shows that defining a module requires `declare module 'xx'`, which I think is required when we write a type definition.

sindresorhus:
I'm gonna let someone with more TS experience chime in, but seems weird to have to define it twice.

sindresorhus:
// @calebboyd @t-sauer

leebenson:
can someone please merge this?

calebboyd:
@u3u There are two ways to declare module types 1) Top Level, and 2) Ambient Modules. 
Chalk is using top level declarations. 
This PR adds an ambient module definition that tries to augment the top level module's definition of the export. Which isn't allowed (is actually a syntax error).

leebenson:
Apologies, I should have actually checked the code before suggesting a PR. I assumed the type syntax would (correctly) be as follows:

```
declare module "chalk" {
  export const enum Level {
    None = 0,
    Basic = 1,
    Ansi256 = 2,
    TrueColor = 3
  }
  
  export interface ChalkOptions {
    enabled?: boolean;
    level?: Level;
  }
  
  export interface ChalkConstructor {
    new (options?: ChalkOptions): Chalk;
    (options?: ChalkOptions): Chalk;
  }
  
  export interface ColorSupport {
    level: Level;
    hasBasic: boolean;
    has256: boolean;
    has16m: boolean;
  }
  
  export interface Chalk {
    (...text: string[]): string;
    constructor: ChalkConstructor;
    enabled: boolean;
    level: Level;
    rgb(r: number, g: number, b: number): this;
    hsl(h: number, s: number, l: number): this;
    hsv(h: number, s: number, v: number): this;
    hwb(h: number, w: number, b: number): this;
    bgHex(color: string): this;
    bgKeyword(color: string): this;
    bgRgb(r: number, g: number, b: number): this;
    bgHsl(h: number, s: number, l: number): this;
    bgHsv(h: number, s: number, v: number): this;
    bgHwb(h: number, w: number, b: number): this;
    hex(color: string): this;
    keyword(color: string): this;
  
    readonly reset: this;
    readonly bold: this;
    readonly dim: this;
    readonly italic: this;
    readonly underline: this;
    readonly inverse: this;
    readonly hidden: this;
    readonly strikethrough: this;
  
    readonly visible: this;
  
    readonly black: this;
    readonly red: this;
    readonly green: this;
    readonly yellow: this;
    readonly blue: this;
    readonly magenta: this;
    readonly cyan: this;
    readonly white: this;
    readonly gray: this;
    readonly grey: this;
    readonly blackBright: this;
    readonly redBright: this;
    readonly greenBright: this;
    readonly yellowBright: this;
    readonly blueBright: this;
    readonly magentaBright: this;
    readonly cyanBright: this;
    readonly whiteBright: this;
  
    readonly bgBlack: this;
    readonly bgRed: this;
    readonly bgGreen: this;
    readonly bgYellow: this;
    readonly bgBlue: this;
    readonly bgMagenta: this;
    readonly bgCyan: this;
    readonly bgWhite: this;
    readonly bgBlackBright: this;
    readonly bgRedBright: this;
    readonly bgGreenBright: this;
    readonly bgYellowBright: this;
    readonly bgBlueBright: this;
    readonly bgMagentaBright: this;
    readonly bgCyanBright: this;
    readonly bgWhiteBright: this;
  }
  
  const chalk: Chalk & { supportsColor: ColorSupport };

  export default chalk;
}
```

calebboyd:
@leebenson In what scenario does `import chalk from "chalk";` not work? The only difference I see with your typings is an Ambient Module declaration which theoretically is no diffrent than a top level export, since those are scoped to the module they were pulled in with (eg. `chalk/index.js`)

PR #247, opened 2018-01-12
Update chalk to use the latest updates to supports-color

partheseas:
supports-color had a breaking change release a few months ago. Instead of assuming stdout as the output stream, it now allows you to check color support on any stream. It does still automatically evaluate stdout, but it is now a subproperty of what the module exports, not the actual export itself.

We can include this update as a non-breaking change in chalk though by simply exporting `chalk.supportsColor` as `supportsColor.stdout`.

Closes #240

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict';
 const escapeStringRegexp = require('escape-string-regexp');
 const ansiStyles = require('ansi-styles');
-const supportsColor = require('supports-color');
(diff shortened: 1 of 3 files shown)


clemtrek:
Duplicate of #240

partheseas:
@clemtrek Except that #240 doesn't pass tests

clemtrek:
Yes... :sob:

sindresorhus:
Looks good. Thanks for your patience :)

PR #252, opened 2018-02-04
Add `wrapper` option

felipenmoura:
The wrapper marks the unprintable characters from each style tag.
This is very useful when debugging or outputing it in different terminals. You can use it to identify the unprintable characters or even to escape them.

Also added tests for it.
Added a description about it to the readme file.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,3 +1,13 @@
+
+## IMPORTANT
+
+This is the same chalk from [](https://github.com/chalk/chalk).  
+The only change made to it is the support for "wrappers" in the effects.
+
+There is a Pull Request opened to add it to Chalk, but while it is not merged (or in case it does not get approved) we can use this fork instead.
+If you want to help, you can check the [Pull Request](https://github.com/chalk/chalk/pull/252) and [the issue](https://github.com/chalk/chalk/issues/253).  
+This is being **temporarily**  published in _npm_ so we can test it out a little further.
+
 <h1 align="center">
 	<br>
 	<br>
@@ -151,7 +161,7 @@ Levels are as follows:
 
 ### chalk.wrapper
 
-The wrapper marks the unprintable characters.  
+The wrapper marks the unprintable characters from style tags.  
 A wrapper can be added to the styles, so you can escape characters or add marks to then.  
 By default, these wrappers are empty strings `""`.
 
--- a/readme.md
+++ b/readme.md
@@ -149,6 +149,26 @@ Levels are as follows:
 2. 256 color support
 3. Truecolor support (16 million colors)
 
+### chalk.wrapper
+
+The wrapper marks the unprintable characters.  
+A wrapper can be added to the styles, so you can escape characters or add marks to then.  
+By default, these wrappers are empty strings `""`.
+
+The wrappers object has two properties, `pre` and `post`.  
+For example:
+
+```js
+const ctx = new chalk.constructor({wrapper: {
+	pre: '>',
+	post: '<',
+}});
+
+ctx.red('foo') // outputs "><foo><"
+```
+
+This can be specially useful when escaping characters, using it into a _PS1_ string or debugging and outputing it into different terminals/TTYs.
+
 ### chalk.supportsColor
 
 Detect whether the terminal [supports color](https://github.com/chalk/supports-color). Used internally and handled for you, but exposed for convenience.
--- a/readme.md
+++ b/readme.md
@@ -165,19 +165,36 @@ The wrapper marks the unprintable characters from style tags.
 A wrapper can be added to the styles, so you can escape characters or add marks to then.  
 By default, these wrappers are empty strings `""`.
 
-The wrappers object has two properties, `pre` and `post`.  
+The wrappers object has two properties, `before` and `after`.  
 For example:
 
 ```js
-const ctx = new chalk.constructor({wrapper: {
-	pre: '>',
-	post: '<',
-}});
+const ctx = new chalk.constructor({
+	wrapper: {
+		before: '>',
+		after: '<'
+	}
+});
 
-ctx.red('foo') // outputs "><foo><"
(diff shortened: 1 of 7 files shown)


readme.md
```diff
@@ -149,6 +149,26 @@ Levels are as follows:
 2. 256 color support
 3. Truecolor support (16 million colors)
 
+### chalk.wrapper
```
sindresorhus: This is not a correct heading as it's not available on the `chalk` global, but is rather a constructor option only.

readme.md
```diff
@@ -149,6 +149,26 @@ Levels are as follows:
...
+```js
+const ctx = new chalk.constructor({wrapper: {
+	pre: '>',
+	post: '<',
+}});
+
+ctx.red('foo') // outputs "><foo><"
+```
+
+This can be specially useful when escaping characters, using it into a _PS1_ string or debugging and outputing it into different terminals/TTYs.
```
sindresorhus: Can you include an example of how it would like for debugging?

readme.md
```diff
@@ -149,6 +149,26 @@ Levels are as follows:
...
+A wrapper can be added to the styles, so you can escape characters or add marks to then.  
+By default, these wrappers are empty strings `""`.
+
+The wrappers object has two properties, `pre` and `post`.  
+For example:
+
+```js
+const ctx = new chalk.constructor({wrapper: {
+	pre: '>',
+	post: '<',
```
sindresorhus: Can you use the PS1 example here instead? For a realistic use-case.

sindresorhus:
I'm ok with adding this as it's a valid use-case and it would be convoluted to add it manually.

@Qix- Thoughts?

sindresorhus:
@felipenmoura It should only be an option in the constructor and not be available on the `chalk` global as a property or as an instance property.

types/index.d.ts
```diff
@@ -11,6 +11,12 @@ export const enum Level {
 export interface ChalkOptions {
 	enabled?: boolean;
 	level?: Level;
+	wrapper?: Wrapper;
+}
+
+export interface Wrapper {
```
sindresorhus: Can you define these inline. Doesn't really make sense to expose a separate `interface` for it.

readme.md
```diff
@@ -149,6 +149,26 @@ Levels are as follows:
 2. 256 color support
 3. Truecolor support (16 million colors)
 
+### chalk.wrapper
+
+The wrapper marks the unprintable characters.  
+A wrapper can be added to the styles, so you can escape characters or add marks to then.  
+By default, these wrappers are empty strings `""`.
+
+The wrappers object has two properties, `pre` and `post`.
```
sindresorhus: Maybe `before` and `after` instead?

sindresorhus:
Can you run the benchmark before and after this change to make sure it doesn't introduce any performance regressions?

Qix-:
I don't see the point in this PR - what is it _really_ trying to solve?

This is what `FORCE_COLOR=1` and `cat -v` (macos) / `cat -A` (linux) / (`man cat` for your arch) are for - so you can debug your output.

felipenmoura:
Hello.
Sure, I will apply the changes and the benchmark tests and let you know.

About the use-case. I have forked chalk and am using it with these changes to build this:

https://github.com/NascHQ/termtools
(It's still in progress, working better on mac)

The point is...if I use chalk to add colours to the output, _PS1_ will use them as expected...but will also count them as "printable". The colour sequences are "unprintable" characters and need to be wrapped between `\[` and `\]`.
So, neither cat nor force-color will prevent that, as I don't want (in this case) to see the characters, but to "cast" them.

If those characters are not seen as "unprintable" by the PS1, they will apply the colours and _look like_ if they had worked, but when the users start typing, their cursor will "go crazy", moving to the wrong positions in the terminal screen (as it uses the number of printable characters to know the cursor position).

I do believe, though, that `cat -v/A` is, indeed, a nice option for debugging.
As for debugging, it would actually offer the option to "tokenize" the colour characters, allowing simpler Regular Expressions for substitutions or counting, for example.

Also useful when we want to truncate a string...if the string has colours, we may show part of that colour code in the result...and also, we would have counted those characters as the "length" of the string.

readme.md
felipenmoura: What header you think I should use? Maybe just "wrapper" ?
Or I could actually add it to the global `chalk`.

felipenmoura:
As for te benchmarks, I'm getting in average this:

**Before:**  

          144,777 op/s » single style
          44,221 op/s » several styles
         260,126 op/s » cached styles
          46,927 op/s » nested styles

**After:**  

         143,500 op/s » single style
          42,878 op/s » several styles
         271,216 op/s » cached styles
          45,659 op/s » nested styles

Also added a few extra benchmarks (demonstrating there might be _some_ difference from when we _do_ use the wrappers, and when we _don't_ use them):

         143,796 op/s » wrapped single style
          42,430 op/s » wrapped several styles
         238,577 op/s » cached wrapped styles
          44,842 op/s » nested wrapped styles

felipenmoura:
Hm, I don't get it.

The PR failed building because:

```
examples/rainbow.js:31:36
  ✖  31:36  Unexpected await inside a loop.  no-await-in-loop
```

But I haven't changed a thing in rainbow.js
Any idea what might have caused this?

Qix-:
@sindresorhus I'm pretty :-1: on this as it's a niche use-case not really suited for chalk. This is what your platform's tools are for and a great use of `FORCE_COLOR`. Unprintable characters are a fact of life, not just in `chalk`, so learning how to debug their output is a pretty essential skill to have.

I don't think introducing _more_ code into `chalk` just for this isolated and ephemeral use-case is valid.

Qix-:
@sindresorhus I'm still of the above opinion. Should we close this?

felipenmoura:
No problem, let's close this and move on.
Thanks for your time.

PR #256, opened 2018-03-03
fix: use export = for typescript module definition

calebboyd:
I actually wasn't sure if this would work but this allows the following in typescript.

```typescript
import * as chalk from 'chalk' // (new versions of ts don't let you use chalk as a function 🙌 )
import chalkDefault from 'chalk'
import chalkTypescriptRequire = require('chalk')// (legacy typescript style)
```
😑 

This will ~fix~ allow intellisense in commonjs projects in vscode (#246)

This may break some people's typescript _type_ usages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
 	"files": [
 		"index.js",
 		"templates.js",
-		"types/index.d.ts"
+		"types/index.d.ts",
+		"types/options.d.ts"
 	],
 	"keywords": [
 		"color",
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,36 +1,14 @@
-// Type definitions for Chalk
-// Definitions by: Thomas Sauer <https://github.com/t-sauer>
-
-export const enum Level {
-	None = 0,
-	Basic = 1,
-	Ansi256 = 2,
-	TrueColor = 3
+/// <reference path="./options.d.ts" />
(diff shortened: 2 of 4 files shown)


calebboyd:
Oh and also re-enables `import { red } from 'chalk'` in ts I know there was some discussion around intent with that...

side note: I wish that the way 99% of commonjs modules were written had a parallel in ES modules 😩. so much `import * as someFunction from 'somewhere'` we'll be "transpiling" forever...

mhegazy:
@sindresorhus, We keep getting user reports running into this issue. is there any thing we can help with to get this issue addressed?

sindresorhus:
> This may break some people's typescript type usages.

Chalk is a very popular package. I'm not sure I want to risk it.

> Oh and also re-enables import { red } from 'chalk' in ts I know there was some discussion around intent with that...

This has already been discussed in great detail over at #215.

Qix-:
What's the verdict here? I'm very out of the loop on typescript stuff so I'm not much help.

If this is an opinionated change that will break existing users then I echo @sindresorhus's [concerns](https://github.com/chalk/chalk/pull/256#issuecomment-384739446).

sindresorhus:
Closing. I'm not interested in risking breaking people's usage for something that is really a bug in VSCode. I also don't want to enable `import { red } from 'chalk'`.

PR #260, opened 2018-03-20
Add Flow type definitions

### LINKED ISSUE #165: Flow type definition
Same as #148, we would be happy to bundle a [Flow](https://flow.org) type definition if anyone wants to do a PR.

saadq:
Closes #165.

It would be great if anyone with Flow knowledge could review this.

And to explain how the test works, basically any comment that has `$ExpectError` in it requires that the line directly below it to have a type error. If there is no type error in the line below `$ExpectError`, the test will fail. The reverse is also true – if there is a type error with no `$ExpectError` above it, the test will fail.

Let me know if there is anything you would like for me to change about this!

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,13 +1,5 @@
 [ignore]
 .*/node_modules/.*
 
-[include]
-
-[libs]
-
-[lints]
-
 [options]
 suppress_comment= \\(.\\|\n\\)*\\$ExpectError
-
-[strict]
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,13 @@
+[ignore]
+.*/node_modules/.*
+
+[include]
+
+[libs]
+
+[lints]
+
+[options]
+suppress_comment= \\(.\\|\n\\)*\\$ExpectError
+
+[strict]
--- a/index.js.flow
+++ b/index.js.flow
@@ -1,0 +1,95 @@
+// @flow
+
+type TemplateStringsArray = $ReadOnlyArray<string>;
+
+export type Level = $Values<{
+	None: 0,
+	Basic: 1,
+	Ansi256: 2,
+	TrueColor: 3
+}>;
+
+export type ChalkOptions = {|
+	enabled?: boolean,
+	level?: Level
+|};
+
+export type ColorSupport = {|
+	level: Level,
+	hasBasic: boolean,
+	has256: boolean,
+	has16m: boolean
+|};
+
+export interface Chalk {
+	(...text: string[]): string,
+	(text: TemplateStringsArray, ...placeholders: string[]): string,
+	constructor(options?: ChalkOptions): Chalk,
+	enabled: boolean,
+	level: Level,
+	rgb(r: number, g: number, b: number): Chalk,
+	hsl(h: number, s: number, l: number): Chalk,
+	hsv(h: number, s: number, v: number): Chalk,
+	hwb(h: number, w: number, b: number): Chalk,
+	bgHex(color: string): Chalk,
+	bgKeyword(color: string): Chalk,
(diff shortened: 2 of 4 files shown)


index.js.flow
```diff
@@ -0,0 +1,95 @@
...
+	has16m: boolean
+|};
+
+export interface Chalk {
+	(...text: string[]): string,
+	(text: TemplateStringsArray, ...placeholders: string[]): string,
+	constructor(options?: ChalkOptions): Chalk,
+	enabled: boolean,
+	level: Level,
+	rgb(r: number, g: number, b: number): Chalk,
```
sindresorhus: Can you not use `this` instead of `Chalk` here? I'm not that familiar with Flow. Just checking. Like: https://github.com/chalk/chalk/blob/58483b5ab110775eb0cfc4999f1cf7e971a00fda/types/index.d.ts#L34

package.json
```diff
@@ -8,14 +8,15 @@
...
+		"test": "xo && tsc --project types && flow --max-warnings 0 && nyc ava",
 		"bench": "matcha benchmark.js",
 		"coveralls": "nyc report --reporter=text-lcov | coveralls"
 	},
 	"files": [
 		"index.js",
 		"templates.js",
-		"types/index.d.ts"
+		"types/index.d.ts",
+		"index.js.flow"
```
sindresorhus: Can you move the file to the `types` directory, same with the Flow test.

.flowconfig
```diff
@@ -0,0 +1,13 @@
+[ignore]
+.*/node_modules/.*
+
+[include]
```
sindresorhus: Can you remove the unused headings like this one?

index.js.flow
jamiebuilds: Not in an interface type https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAdgFwKYBOUAhgMZ5gDCAFsTANZgDeqYYBA5gEYAUBAXGCwBXALZdCAGjAdBI8VLBc5YiQQCUgnNQwBnANyoAvuiA

.flowconfig
saadq: Done!

package.json
saadq: I wish this was possible :/

Originally, I had wanted this PR to have a folder structure like so:

```
types
├── flow
│   ├── .flowconfig
│   ├── index.js.flow
│   └── test.js
└── typescript
    ├── index.d.ts
    ├── test.ts
    └── tsconfig.json
```

There are some limitations on the Flow side of things though. 

1. The `.flowconfig` needs to be within the root project folder that has `node_modules` in it. We could get around this by having a script like this though:

```json
{
  "scripts": {
    "flow": "cd types/flow && flow --max-warnings 0"
  }
}
```

2. The `index.js.flow` file needs to be in the same folder of the file it is shadowing. Because there is no build process in your project (where every file ends up in `lib` for example), I don't think there is any way to get around this, so the only way I can think of to allow this is to introduce a build process in your project which creates a `lib` folder that has `index.js`, `index.d.ts`, and a`index.js.flow` inside it. I don't know if you'd be willing to do something like that for this change though.

@jamiebuilds or anyone else please correct me if you have any better suggestions.

saadq: Also note: there is an issue open [here](https://github.com/facebook/flow/issues/2675) for Flow that would allow you to have a field in `package.json` to point to the flow libdef, but currently there is nothing equivalent to TypeScript's `types`/`typings` field :(

sindresorhus: I see. Ok, what you have is good for now then. We can improve it in the future when/if Flow adds the package.json field.

sindresorhus:
Thank you, @saadq! We appreciate your work on this 🙌

PR #265, opened 2018-04-22
[flow] add compatibility with node require()

### LINKED ISSUE #264: Flow types don't work when using `require`
When using require:

```javascript
```

The flow types for `chalk` says:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/commands/login.js:34:42

Cannot call chalk.red because property red is missing in module chalk [1].

 [1]  4│ const chalk = require('chalk');
       :
     31│   if(!team) {
     32│     console.log('Failed to login')
     33│   } else {
     34│     console.log('Logged in as %s', chalk.red(team.name))
     35│   }
     36│
     37│   close()
```

and that's because the flow types expect you to explicitly require `default`. i.e.

```javascript
const chalk = require('chalk').default;
```

and only then does it work correctly. i.e. it seems like `index.flow.js` was built to support `import` but not `require`. @saadq could you take a look?

saadq:
Fixes #264. I forgot to take into account that not everyone who uses Flow uses ES Modules, so using a CommonJS export will you to use both:

```js
```

as well as 

```js
import chalk from 'chalk';
```

It is fine to leave the `export` on the types however. Regardless if you're using CJS or ESM, you will be able to use `import type` syntax such as:

```ts
import type { Level } from 'chalk';
```

This will work even if you're using `require` because that flow import syntax is just stripped away with Babel just like the rest of the types.

I didn't add an additional test file which uses the `require` syntax, but I tested it locally and it worked. If you want me to add a `_flow-node.js` test file or something let me know and I can add that as well.

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -90,6 +90,4 @@ export interface Chalk {
 	supportsColor: ColorSupport
 };
 
-declare var chalk: Chalk;
-
-export default chalk;
+declare module.exports: Chalk;


pranaygp:
Ohh damn, I just made #266 to fix this too but didn't see your PR in time 😝

saadq:
@pranaygp Since you're a first-time contributor here and the fact that your version actually makes more sense – (I'm not actually sure why I switched to a declaration syntax when using the CJS version, although both are fine), I'm going to close this and we can favor your PR instead.

If Sindre asks for another test file, you can just create a file in `test` called `_flow-node.js` or something and just copy all of `_flow.js` into it and just change the `import` syntax to `require`.

PR #266, opened 2018-04-22
[flow] export default -> module.exports (to support require and import)

### LINKED ISSUE #264: Flow types don't work when using `require`

pranaygp:
Fixes #264 

This seems to not fail the existing tests and it works when I test it with require, however I'm not 💯% on this, so someone who knows flow better could check this?

@saadq?

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -91,5 +91,4 @@ export interface Chalk {
 };
 
 declare var chalk: Chalk;
-
-export default chalk;
+module.exports = chalk;


saadq:
👌 from me

sindresorhus:
Closing in favor of #268

PR #286, opened 2018-07-22
Add type definitions badge

paulmelnikow:
Hi @sindresorhus! I added this badge to Shields at your suggestion badges/shields#1252, and though I'd see if you'd like to include it here!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,7 +9,7 @@
 
 > Terminal string styling done right
 
-[![Build Status](https://travis-ci.org/chalk/chalk.svg?branch=master)](https://travis-ci.org/chalk/chalk) [![Coverage Status](https://coveralls.io/repos/github/chalk/chalk/badge.svg?branch=master)](https://coveralls.io/github/chalk/chalk?branch=master) [![](https://img.shields.io/badge/unicorn-approved-ff69b4.svg)](https://www.youtube.com/watch?v=9auOCbH5Ns4) [![XO code style](https://img.shields.io/badge/code_style-XO-5ed9c7.svg)](https://github.com/xojs/xo) [![Mentioned in Awesome Node.js](https://awesome.re/mentioned-badge.svg)](https://github.com/sindresorhus/awesome-nodejs)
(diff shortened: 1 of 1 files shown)


paulmelnikow:
It _is_ really long.

The reason I had included the version numbers was to help developers distinguish between well maintained libraries, whose defs are being kept up to date, and libraries which aren't. It seems like the version number is useful for that. That said, I'm relatively new to TypeScript and Flow so I appreciate your perspective on this!

1. 
2. 
3. 
4. 
5. 

Further thoughts?

sindresorhus:
> The reason I had included the version numbers was to help developers distinguish between well maintained libraries, whose defs are being kept up to date, and libraries which aren't. It seems like the version number is useful for that.

What if my library is mature and stable and doesn't need updates? Many packages don't even have TS or Flow as development dependencies.

---

I would go for number 2.

paulmelnikow:
I see. Thanks. :+1:

Could I ask a couple other questions?

For packages that don’t declare the dependencies, is there a better way to infer their type support?

Is it possible that a future breaking change to the language would require definitions to be updated, even if the library is mature?

sindresorhus:
> For packages that don’t declare the dependencies, is there a better way to infer their type support?

For TS, you could look for a `index.d.ts` file in root or a `types` field in package.json.

For Flow, you could look for a `index.js.flow` file in root or a `flow-typed` directory in root.

> Is it possible that a future breaking change to the language would require definitions to be updated, even if the library is mature?

Yes, but I don't think a version is a good enough indicator of that. TS might break a library in a minor release or never. It really depends on what features you use and how. For simple node modules, it's unlikely new versions will break much.

paulmelnikow:
Great. Thanks so much for the feedback. Give me a few days and I will make some updates to the badge.

sindresorhus:
@paulmelnikow Can this be finished now?

Qix-:
Late to the game but 2 was the choice I would have made as well.

Just for some validation 💃

Qix-:
Resolved the conflicts. Only issue I see, which is super nit-picky, is a wrap of a single badge:

image

Qix-:
SGTM, thanks @sindresorhus and @paulmelnikow!

paulmelnikow:
Ah yep! Thanks for merging this!

PR #291, opened 2018-08-05
Serve logo with proper MIME type

balcsida:
> Chalk logo on npm.com: I don't feel so good, Mr. Stark.

> Strange: I looked into the public npm repository. I saw 761,349 packages.
> Stark: How many of those had incorrect incorrectly served images in their readme?
> Strange: Many of them. But Chalk was the most popular repo of all of them

Sorry for the above one, I did this PR 1:35 AM and wanted to have some extra fun.
If you would like to use rawgit as you used in the screenshot, please let me know and I can use that.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 <h1 align="center">
 	<br>
 	<br>
-	<img width="320" src="media/logo.svg" alt="Chalk">
+	<img width="320" src="media/logo.svg?sanitize=true" alt="Chalk">
 	<br>
 	<br>
 	<br>


Qix-:
Can you explain the use of this query param?

sindresorhus:
Report it here: https://github.com/npm/marky-markdown/issues I'm not interested in changing something to work around npm's lousy Markdown rendering :)

balcsida:
@Qix- : it serves the linked file with an appropriate content type, in this case, instead of `text/plain` it will be `image/svg+xml`
Please check the difference here:
https://raw.githubusercontent.com/chalk/chalk/master/media/logo.svg
https://raw.githubusercontent.com/chalk/chalk/master/media/logo.svg?sanitize=true

@sindresorhus With due respect, I don't think this is the issue of npm's markdown rendering. GitHub uses this parameter when displaying an image.

sindresorhus:
> With due respect, I don't think this is the issue of npm's markdown rendering. GitHub uses this parameter when displaying an image.

It's GitHub that adds that parameter and so should npm. If npm wants to render GitHub readme's they should do it properly and follow how GitHub does it.

sindresorhus:
> npm's lousy Markdown rendering

For context, this is far from the first npm Markdown issue I've had to deal with. Probably the twentieth or more issue. So excuse my grumpiness 😃🙌

PR #296, opened 2018-09-05
Remove const enum from Level TypeScript definition (#295)

### LINKED ISSUE #295: TypeScript error when using --isolatedModules
Hi! Sorry to bother you, but your type definitions seem to be throwing an error when compiling TS:

```
$ tsc
node_modules/chalk/types/index.d.ts:4:19 - error TS1209: Ambient const
enums are not allowed when the '--isolatedModules' flag is provided.

export const enum Level {
                  ~~~~~
```

strothj:
Fixes #295.

I changed the type definition for `Level` to use a naked enum. This makes the type definitions compatible with Babel compiled TypeScript.

I enabled the `isolatedModules` compiler flag so that the introduction of incompatible code will be caught as an error.

`const enum`, if I'm not mistaken, only controls emit behavior. Changing it to a normal `enum` in the type definitions should not change the semantic meaning.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,7 +1,7 @@
 // Type definitions for Chalk
 // Definitions by: Thomas Sauer <https://github.com/t-sauer>
 
-export const enum Level {
+export enum Level {
 	None = 0,
 	Basic = 1,
 	Ansi256 = 2,
--- a/types/tsconfig.json
+++ b/types/tsconfig.json
@@ -4,6 +4,7 @@
 		"target": "es6",
 		"noImplicitAny": true,
 		"noEmit": true,
-		"allowJs": true
+		"allowJs": true,
+		"isolatedModules": true
 	}
 }


sindresorhus:
[`const enum`](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) is a commonly used language feature. We're not going to change just because Babel doesn't fully support TS. It's up to Babel to support the TS ecosystem, not the other way around.

sindresorhus:
You should open an issue on Babel instead.

strothj:
Sure, no problem. Thank you for your time.

chicoxyzzy:
Actually it's not Babel issue. It's more like TypeScript issue with isolated modules.

the-simian:
There appears to be some discussion on this here: https://github.com/Microsoft/TypeScript/issues/20703

Kind of a bummer, but I'm using create-react-app which (if you don't eject) heavy-handedly forces isolatedModules to true. I ejected fro months and months, but I'm trying to leverage the benefits of not ejecting at this time. That means that the only workaround I have, as I am beholden to chalk for a 3rd party dependency, is to set `skipLibCheck` to `true` in the tsconfig.json. This works, with the obvious downsides.

alangpierce:
@the-simian I submitted https://github.com/Microsoft/TypeScript/pull/28465 and I *think* (but haven't confirmed) that it will be released in TypeScript 3.4, coming out very soon. That means that after upgrading TypeScript, you'll only get errors if you try to use `Level` enum, not by simply including chalk as a dependency.

(For other reasons mentioned above, I still think it's best for libraries to either ship concrete (non-`const`) enums or use a union type instead.)

Macil:
@alangpierce On Typescript 3.7.5, I get this error just from `import 'chalk'` without using anything from it:

```
node_modules/chalk/index.d.ts:403:16 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.

403  Level: typeof LevelEnum;
                   ~~~~~~~~~


Found 1 error.
```

I wonder if that change to typescript was not enough, or if the issue was re-introduced in either typescript or by some change in this library since then.

strothj:
> 
> 
> @alangpierce On Typescript 3.7.5, I get this error just from `import 'chalk'` without using anything from it:
> 
> ```
> node_modules/chalk/index.d.ts:403:16 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.
> 
> 403  Level: typeof LevelEnum;
>                    ~~~~~~~~~
> 
> 
> Found 1 error.
> ```
> 
> I wonder if that change to typescript was not enough, or if the issue was re-introduced in either typescript or by some change in this library since then.

That error is caused by the fact that you have the `compilerOption` `isolatedModules` set to true. This is normally set to true when using something like Babel to compile your code. `isolatedModules` means that each module needs to be able to be compiled in isolation. The `const enum` cannot.

You can add `"skipLibCheck": true,` to your `compilerOptions` to disable type checking of type declarations.

Issue #300, opened 2018-09-18
Roadmap ideas


sindresorhus:
Here are the goals for Chalk 3:

- [x] Drop support for Node.js 4
- [ ] Bundle all dependencies at publish-time. I would usually never do this, but Chalk is one of the most popular packages on npm. We could use Browserify or Webpack something to make it a single file.
- [ ] `require('chalk/lite')` which only includes support for 16-bit. For when you need something fast and simple for logging.
- [x] Move to class-based construction: #188 
- [x] Improve speed, both require speed (#180) and execution speed (https://github.com/chalk/chalk/issues/333).
	For execution speed we could add fast-paths for when only using 16 colors, which is the most common use-case.
- [ ] Stop using the `color-convert` module. It's too slow and flexible. We don't really need that many ways of specifying colors. I would prefer to support *just* keywords, HEX, RGB. I'm honestly not sure if this is feasible. Just an idea for improving the package size and performance. Alternatively, make `color-convert` tree-shakable, so we only get the parts we actually need.
- [ ] Browser DevTools support built-in? This has been a common request and as long as it doesn't require too much code (and no new dependencies), I welcome a PR adding it.

**We welcome contributions to help with this effort and ideas for the v3 goals. 🙌**

---

@Qix- Thoughts? Anything else?

sindresorhus:
Sure, Rollup would work too.

Qix-:
Sorry @sindresorhus, commented too quickly. Edited my comment with more information.

pspeter3:
Is anyone owning the TypeScript conversion? I would be happy to help with that and bundling during publish

Qix-:
@pspeter3 go for it - pull requests are always welcomed.

felixfbecker:
I dont really understand the benefit of bundling. Why can’t the consumer bundle? If the packages bundle, consumers lose the ability to dedupe, which makes the end bundle larger, not smaller

jaylinski:
Another goal could be to reduce the install size: https://packagephobia.now.sh/result?p=chalk

(Although it's already pretty good.)

Qix-:
@jaylinski 90 kilobytes is not a lot. I'm not worried about that, really.

@styfle good catch, thanks for submitting that PR to them.

[...]

PR #304, opened 2018-10-02
Remove dependency on escape-string-regexp

Chalarangelo:
As `chalk` is used by thousands of people daily and provided that `escape-string-regexp` is a very simple package, I went ahead and removed the extra dependency from `package.json`, replacing it with an inline function. Feel free to close, if you believe this could introduce problems down the line, as it is an opinionated change.


sindresorhus:
No thanks, but you should take a look at https://github.com/chalk/chalk/issues/300 where I propose a better solution ;)

PR #309, opened 2018-11-25
Strict mode in Flow definition

yungsters:
Converts the Flow definition for `chalk` to use `@flow strict`.

This makes it so that projects using `chalk` can use `@flow strict` (which requires all dependencies to be recursively strict).

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -1,4 +1,4 @@
-// @flow
+// @flow strict
 
 type TemplateStringsArray = $ReadOnlyArray<string>;
 


sindresorhus:
I assume this change doesn't break Flow projects that does not use the Flow strict mode?

nvie:
Correct! It just makes a stricter guarantee for projects that want it, without breaking anything for projects that don't require strictness.

sindresorhus:
Done: https://github.com/chalk/chalk/releases/tag/v2.4.2

nvie:
🙌

sindresorhus:
FYI: https://github.com/chalk/chalk/commit/d3be9c65b11f505e0e8bd9d4796c4988bb634cf7

Would be great if either of you could submit it to https://github.com/flow-typed/flow-typed

PR #313, opened 2018-12-11
fix(ts): normalize default commonjs export

Hotell:
currently when using chalk with typescript or `//@ts-check` within vanilla js, user has to import `default` which is indeed very strange and non standard api. 

This PR fixes that issue for both vanilla JS users using TS to check js files and TS users

*Before:*

```js
// @ts-check
const {default:chalk} = require('chalk')
```

*After:*
```js
// @ts-check
const chalk = require('chalk')
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -222,4 +222,3 @@ Object.defineProperties(Chalk.prototype, styles);
 
 module.exports = Chalk(); // eslint-disable-line new-cap
 module.exports.supportsColor = stdoutColor;
-module.exports.default = module.exports; // For TypeScript
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,31 +1,31 @@
 // Type definitions for Chalk
 // Definitions by: Thomas Sauer <https://github.com/t-sauer>
 
-export const enum Level {
+declare enum Level {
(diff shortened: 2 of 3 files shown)


Qix-:
14 hours and waiting to hear back from testing 🤔

Hotell:
well, Travis "rocks" again 🗡

Qix-:
I think I got the all-clear to move chalk to circle at one point. Might do that.

sindresorhus:
The current implementation is intentional. There's no way to correctly support ES2015 and CommonJS at once, and not interested in using `export =` type of syntax, so we have opted for optimizing for ES2015. Just use `default` if you want to use it in CommonJS.

Hotell:
> There's no way to correctly support ES2015 and CommonJS at once

that's not true and that's what's this PR demonstrates, anyway you've other reasons. Well 🤷‍♂️ .

Anyways I found better package which solves things correctly, if anyone else reading this comes into same issue https://github.com/lukeed/kleur

cheers

ma2ciek:
Thanks, @Hotell. The mentioned library works for me better.

PR #316, opened 2018-12-23
Change tagged template literal argument type to accept `unknown` instead of just `string`

### LINKED ISSUE #314: Incorrect types for chalk in tagged template literal form
I am using chalk to do some logging and I noticed this issue.
```
chalk`{black.bold  Worker:${process.pid}: } {blue.bold job received}: ${job.id}`
```
In this case I am giving chalk two numbers instead of strings.
Typescript complains about this because chalk's tagged template literal form is defined as 
```
```
In my case placeholders contains numbers. 

Now I could fix this by doing  
```
chalk`{black.bold  Worker:${`${process.pid}`}: } {blue.bold job received}: ${`${job.id}`}`
```
but that feels wrong.

Ideally it would be some type union of things easily parsed to string (<string>|<number>|etc etc)
```
(text: TemplateStringsArray, ...placeholders: ParseableToString[]): string;
```

vanbujm:
Fixes #314. Changed Type to `unknown` as per @sindresorhus's request.


vanbujm:
One thing to be aware of is `unknown` will break types for Typescript 2.x versions. Happy to submit a more complicated type def that uses any for older versions and unknown for newer versions if we want to support older TS versions.

sindresorhus:
> One thing to be aware of is unknown will break types for Typescript 2.x versions.

That's fine. This change will be part of the next major release anyway.

types/index.d.ts
```diff
@@ -27,7 +27,7 @@ export interface ColorSupport {
 
 export interface Chalk {
 	(...text: string[]): string;
```
sindresorhus: This should be changed too.

```suggestion
	(...text: unknown[]): string;
```

We coerce this argument to string: https://github.com/chalk/chalk/blob/076f0c9eb6d6575db745f2753832918eaefe627d/index.js#L167

sindresorhus:
Can you also update the Flow definition?

vanbujm:
@sindresorhus, I had a crack at changing the types for flow.

However there are two flow tests that check for null, that fail now since a `mixed[]` can be null.
I deleted them since it is no longer a typing error. But I'm unsure if that is the right call to make.

sindresorhus:
Can you fix the merge conflict?

sindresorhus:
> However there are two flow tests that check for null, that fail now since a mixed[] can be null.
I deleted them since it is no longer a typing error. But I'm unsure if that is the right call to make.

All good 👍

sindresorhus:
Looks like you resolved the merge conflict incorrectly. Look at the diff.

vanbujm:
@sindresorhus Okay, I think it is sane now. I rebased when I did the first merge and thats what made everything crazy. Ended up just going back to pre-merge, merging (correctly) this time, then force pushing my fork. There was probably more elegant git wizardry that could have been done but ¯\_(ツ)_/¯

PR #323, opened 2019-01-26
add chalker to related

jchip:
I created a simple module for using chalk with strings like `<red>red text</>`.  Would you accept this PR to add it to related in readme in case anyone find it useful please?  thanks.


jchip:
Mainly for tooling and preference.  XML is just the `<>` and `</>` marker convention, like HTML is XML.  Many editors have automatic support for editing these, even within strings.

Qix-:
HTML is not XML, just to be clear.

I'll leave this up to @sindresorhus.

jchip:
Thanks for your time.  I am closing this to avoid taking any more of your time.  Thanks.

PR #328, opened 2019-03-01
Update depended packages count

Richienb:
Updated the number of packages depended on.


Qix-:
Amazing :) thank you!

PR #330, opened 2019-03-12
Remove the blue color workaround for Windows

### LINKED ISSUE #329: Give some method to use original "blue" even under Windows
I saw this in `index.js`

```js
// Use bright blue on Windows as the normal blue color is illegible
if (isSimpleWindowsTerm) {
	ansiStyles.blue.open = '\u001B[94m';
}
```

But sometimes we may need to explicitly use the original `blue` even if it is not legible.

sindresorhus:
The illegible blue color has been fixed in Windows 10 build 16257: https://blogs.msdn.microsoft.com/commandline/2017/08/02/updating-the-windows-console-colors/

The workaround causes all kinds of problems, so better to remove it than adding conditionals for older Windows versions.

Fixes #329

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,11 +45,6 @@ function Chalk(options) {
 	applyOptions(this, options);
 }
 
-// Use bright blue on Windows as the normal blue color is illegible
-if (isSimpleWindowsTerm) {
-	ansiStyles.blue.open = '\u001B[94m';
-}
-
 for (const key of Object.keys(ansiStyles)) {
 	ansiStyles[key].closeRe = new RegExp(escapeStringRegexp(ansiStyles[key].close), 'g');
 
--- a/test/windows.js
+++ b/test/windows.js
@@ -29,18 +29,6 @@ test.beforeEach(() => {
 test('detect a simple term if TERM isn\'t set', t => {
 	delete process.env.TERM;
 	const chalk = importFresh('..');
-	t.is(chalk.blue('foo'), '\u001B[94mfoo\u001B[39m');
-});
-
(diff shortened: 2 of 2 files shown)


sindresorhus:
@be5invis Do you know if https://github.com/chalk/chalk/issues/58 has been fixed in Windows 10 yet? Would have been nice to be able to remove https://github.com/chalk/chalk/blob/60959e05cfe7086a38927eae5b5e04c459ce4182/index.js#L144-L145 too.

be5invis:
@sindresorhus 
This is in RS5 (1809)
image
The workaround may be still needed for legacy versions (7, 8, etc.). Perhaps you can still enable it when `supports-color` returns 1.

sindresorhus:
@be5invis #58 was using cmd.exe, not PowerShell.

be5invis:
@sindresorhus 
image
I think this is more clear.
I do not have older builds currently (though there server have **all** the historical Windows builds).
You can remove these workarounds only for version >= 17763.

Qix-:
Have they also fixed the dim issue entirely? If so then let's remove the simple windows term check entirely as @be5invis mentioned.

sindresorhus:
> Have they also fixed the dim issue entirely? If so then let's remove the simple windows term check entirely as @be5invis mentioned.

Seems so from the comments above. I'll do that in a separate PR.

sindresorhus:
> You can remove these workarounds only for version >= 17763.

Not interested in supporting older builds. This workaround complicates a lot of code. Users can easily upgrade to latest Windows builds.

sindresorhus:
https://github.com/chalk/chalk/pull/331

PR #335, opened 2019-03-21
Fix support for nested styles

### LINKED ISSUE #334: Nesting bug
[Issuehunt badges](https://issuehunt.io/r/chalk/chalk/issues/334)

If the bug mention in https://github.com/doowb/ansi-colors#nested-styling-bug is valid, we should fix it.

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/chalk/chalk/issues/334).

farnabaz:
As I recovered both `bold` and `dim` style has same closing chars and this cause the problem
```
bold: { open: '\u001b[1m', close: '\u001b[22m' }
dim: { open: '\u001b[2m', close: '\u001b[22m' }
```
On line 165 all closing chars on current style replace by opening chars, so in this specific situation closing tag of `dim` replaced by `bold` opening tag and `bar` seen as `dim` 
https://github.com/chalk/chalk/blob/7b9211be501f0608ed89c1a6da88a144199b7a54/index.js#L165

Fixes #334


Qix-:
Unfortunately this doesn't solve it. This is intentional. The history of escape codes is annoying and unclear, but `22m` is the only widely compatible way to close either bold or dim. `21m` is not used as frequently since it has been confused as a double-underline, which isn't commonly supported in rendering libraries and thus is usually just silently ignored.

I wish it was this easy.

farnabaz:
Thank you @Qix- for information. I didn't know much about escape code, I'have checked `ansi-colors` and compared its result with `chalk`:  
```bash
'\u001b[1mfoo \u001b[31m\u001b[2mbar\u001b[22m\u001b[1m\u001b[39m baz\u001b[22m' ansi-colors
'\u001b[1mfoo \u001b[31m\u001b[2mbar          \u001b[1m\u001b[39m baz\u001b[22m' chalk
```
As you can see the only difference between `chalk` and `ansi-colors` is `22m` that closes `dim`, So I thought I found problem, I've change line 165 to prevent removing closeing char of `dim` and it works. 

Also all test passed except the one that compare nesting style result. 
Here is result:
Screen Shot 2019-03-21 at 10 12 18 PM

mirefly:
This is how `ansi-colors` solve the nested styling bug. To me, it looks like a good way to do it without introducing a global.

Qix-:
Yeah, this seems right. @sindresorhus at some point we need to condense codes down into a single CSI character and not repeat it because that output is ugly (not your problem @farnabaz).

sindresorhus:
@mirefly Can you fix the merge conflict?

sindresorhus:
Can you also try to write a more descriptive PR title of what this fixes?

mirefly:
@sindresorhus  I guess you mean the author @farnabaz ?

sindresorhus:
Yes, sorry.

farnabaz:
@sindresorhus Done

PR #336, opened 2019-03-24
Export Color type

natealcedo:
The PR exposes a type for colors supported by chalk. The use case of this is so that consumers of this library can programatically check if a color is available. 

One example is in this PR. https://github.com/facebook/jest/pull/8025

I am hardcoding the possible values that a color can be. https://github.com/facebook/jest/pull/8025/files#diff-6170a10d44dcb3222b64fba6427d69c7R234

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -67,6 +67,45 @@ export interface ColorSupport {
 	has16m: boolean;
 }
 
+/**
+ * Available colors for use
+ */
+export type Color =
+	"black"
+	| "red"
+	| "green"
+	| "yellow"
+	| "blue"
+	| "magenta"
+	| "cyan"
+	| "white"
+	| "gray"
+	| "grey"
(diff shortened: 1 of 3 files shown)


sindresorhus:
Makes sense, but I think we should split it in `ForegroundColor`, `BackgroundColor`, and `Color`, which is a combination of the two. Maybe we should also expose the modifiers too?

SimenB:
@natealcedo ping 🙂

natealcedo:
Hello again! Sorry I neglected this. Thanks @SimenB for reminding me. Let me take a look at this again over the weekend :)

natealcedo:
Hi actually, I've been terribly busy with work and my personal life. I'm not sure if I can pick this up anytime soon if anyone wants to take a stab at this. :)

SimenB:
Opened up #357

PR #342, opened 2019-04-25
Fix TypeScript types for `supportsColor` which is top‑level only

ExE-Boss:
The `supportsColor` property only exists on the top level:
```js
chalk.supportsColor; // object
chalk.bold.supportsColor; // undefined

'supportsColor' in chalk; // true
'supportsColor' in chalk.bold; // false
```

review?(@BendingBender, @Qix-, @sindresorhus)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -70,11 +70,25 @@ declare namespace chalk {
 		has16m: boolean;
 	}
 
-	type ChalkTemplateFunction = (text: TemplateStringsArray, ...placeholders: unknown[]) => string;
+	interface ChalkFunction {
+		/**
+		Use a template string.
 
-	interface Chalk {
+		@remarks Template literals are unsupported for nested calls (see [issue #341](https://github.com/chalk/chalk/issues/341))
+
+		@example
+		log(chalk`
+		CPU: {red ${cpu.totalPercent}%}
+		RAM: {green ${ram.used / ram.total * 100}%}
+		DISK: {rgb(255,131,0) ${disk.used / disk.total * 100}%}
+		`);
+		*/
+		(text: TemplateStringsArray, ...placeholders: unknown[]): string;
+
 		(...text: unknown[]): string;
+	}
 
+	interface Chalk extends ChalkFunction {
 		/**
 		Return a new Chalk instance.
 		*/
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -99,11 +99,9 @@ declare namespace chalk {
 		@param color - Hexadecimal value representing the desired color.
 
 		@example
-		```
```
ExE-Boss: These are unnecessary as `@example` already creates a code block and using these causes the code example to be indented unnecessarily in **VS Code’s** IntelliSense popout.

BendingBender: This conforms with @sindresorhus's [TS styleguide](https://github.com/sindresorhus/typescript-definition-style-guide).

index.d.ts
```diff
@@ -273,13 +279,15 @@ declare namespace chalk {
 	}
 }
 
+declare type NoThis<T> = {[K in keyof T]: T[K] extends T ? T : T[K]};
```
BendingBender: Can you please elaborate on this?

As far as I understand it, this will take all properties from `T` that subclass `T` and replace them with `T`. What properties/use case does this target?

ExE-Boss: That without this, TypeScript considers the following valid:
```ts
import chalk = require('chalk');

// Will throw at runtime because `chalk.reset.supportsColor` is `undefined`:
chalk.reset.supportsColor.level;
```

This is because without this, TypeScript resolves `chalk.reset` to `typeof chalk` rather than `Chalk`.

index.d.ts
ExE-Boss: Well, then @sindresorhus’s style guide should be amended.

index.d.ts
BendingBender: Why do we keep returning `this` then? Wouldn't it be more straightforward to simply return the `Chalk` interface instead of `this` (given `Chalk` and `ChalkFunction` are merged together)?

ExE-Boss: Well, unless we were to use proxies to have the style builder properties return the object from which they’re requested, then yes, it’s probably more straightforward to just return `Chalk`.

Also, the following doesn’t work for whatever reason:
```js
class CustomChalk extends chalk.Instance {
	constructor(options) {
		super(options);
	}
}

new CustomChalk(); // Is of type Chalk and not CustomChalk
```

Also, `instanceof` checks don’t work either.

BendingBender: The subclassing issue is a valid one, is this something that makes sense at all for Chalk? Any use cases? @sindresorhus

Qix-: IIRC the subclassing issue was purely for the case of attaching your own colors to chalk.

index.d.ts
sindresorhus: Without them VS Code's popout doesn't preserve tab-indentation. I'm also not really interested in working around VS Code issues. Open a VS Code issue about this instead. I'm not going to change the style guide about this.

sindresorhus:
Thanks :)

PR #343, opened 2019-04-27
fix: Add Babel compatible `__esModule` export

ExE-Boss:
Since this module now supports the `default` export, it makes sense to add the Babel compatible `__esModule` export.

## See also:
- https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs

### DIFF
--- a/index.js
+++ b/index.js
@@ -197,6 +197,9 @@ const chalkTag = (chalk, ...strings) => {
 
 Object.defineProperties(Chalk.prototype, styles);
 
-module.exports = Chalk(); // eslint-disable-line new-cap
-module.exports.supportsColor = stdoutColor;
-module.exports.default = module.exports; // For TypeScript
+exports = module.exports = Chalk(); // eslint-disable-line new-cap, no-multi-assign
+exports.supportsColor = stdoutColor;
+
+// For TypeScript and Babel:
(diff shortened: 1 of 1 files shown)


Qix-:
I've literally never had a problem with this. What problem is this solving?

ExE-Boss:
It tells Babel and TypeScript that this module has an ES6‑compatible `default` export, which makes the `import` transpiler not emit useless object creation.

---

```js
import foo from 'foo';
foo('stuff');
```
gets transpiled as:
```js
/** @type {(<T extends {default: any}>(mod: T) => T) & (<T>(mod: T) => {readonly default: T})} */
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
var foo = __importDefault(require('foo'));
foo.default('stuff');
```

When `foo` doesn’t have `__esModule`, then every time it’s imported in transpiled code, a new object containing a single `default` property will be created.

When it does have `__esModule`, then it will be returned directly.

Qix-:
Up to you @sindresorhus. The boilerplate creeps more and more each year it seems.

sindresorhus:
See https://github.com/sindresorhus/ora/pull/111#issuecomment-487273101

PR #344, opened 2019-04-27
Use CommonJS-compatible export in TypeScript definition

BendingBender:
See https://github.com/sindresorhus/meta/issues/9#issuecomment-487273390.


sindresorhus:
Thanks for doing this 🙌

PR #317, opened 2018-12-26
Add `stderr` support

langpavel:
This add support for `require('chalk/stderr')`

Fixes #301 
Fixes #318

- [x] tests
- [x] documentation (may need rewording)
- [x] typescript definition (may be incorrect), see comment
- [x] flow definition

PR #354, opened 2019-07-12
Add support for chalk/stderr

### LINKED ISSUE #301: Auto-detect support for `stderr`
[Issuehunt badges](https://issuehunt.io/r/chalk/chalk/issues/301)

I want use colors only on `console.error` calls which is typically terminal.
I suggest some API extension, very easy idea is chaining like this:
```js
const coloredOutput = chalk.stderr.red('RED ALERT');
```
This can be used more easily like this:
```js
const chalk = require('chalk').stderr;
const coloredOutput = chalk.red('RED ALERT');
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)
### Submitted pull Requests
- [#359 Add chalk.stderr and chalk.stderr.supportsColor](https://issuehunt.io/r/chalk/chalk/pull/359)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/chalk/chalk/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

Qix-:
Closes #301.

The diff is annoyingly misleading so here's exactly what happened:

1. Remove `supports-color` from `source/index.js` in lieu of a generic factory function
2. Remove the unnecessary `Chalk` function and references to `Chalk.prototype` (a regular object works fine instead).
3. Export simply `chalkFactory` in `source/index.js instead of the nicely wrapped `chalk` instance
4. Rename `source/index.js` to `source/factory.js`, which becomes the basis for both stderr and stdout instances
5. Create `source/index.js` and `stderr.js`, both of which are small glue wrappers around the factory function (virtually the identical code removed from the now-renamed `source/index.js`) that also pass their respective `supports-color` objects.

### Caveat 1: npm

There's no seemingly elegant way to be able to have `require('chalk/stderr')` pull from `source/stderr.js` so I pulled `stderr.js` into the root. It might make sense to also pull `index.js` into the root for consistency, which would allow us to remove `main` from package.json. It might also make sense to rename `source` -> `lib` in that case, too.

Seems like a giant step back but unfortunately npm is very limiting when it comes to this kind of thing.

### Caveat 2: Typescript

Typescript type definition files are a nightmare, I couldn't figure out how to get typescript to recognize them. Perhaps @blakeembrey can help - as is, types do not import correctly using a tsserver instance even before these changes (on master) so something was broken beforehand.

Unless I'm mistaken, only an `stderr.d.ts` would have to be created with a `///<reference path="./index.d.ts" />` at the very top since the returned "types" from both `index.js` and `stderr.js` are the same.

However, none of that seems to be working. I have no idea how to debug the issue, so I'm keeping this PR strictly changes to the functional code. We should investigate typescript types before releasing chalk 3.


blakeembrey:
@Qix- Do you want to check in what you did for `stderr.d.ts` or would you like me to take a pass at it in a PR to this PR? You don't actually want any reference to it, but you do need to `import` and re-`export` the types you want to use from `index.d.ts`. You could also put those in a third file depending on preferences.

Qix-:
@blakeembrey A PR would be nice, though I'm still confused as to why tsserver doesn't pick the types up. IIRC (not in front of a machine at the moment) even tsc was unable to find the types.

sindresorhus:
@Qix- This is not what we decided on in https://github.com/chalk/chalk/pull/317#issuecomment-471996846

Qix-:
You're right, not sure how I got my signals crossed. I can try to fix it up but feel free to merge the other one if it works better.

sindresorhus:
Closing in favor of https://github.com/chalk/chalk/pull/359.

PR #356, opened 2019-07-12
Remove the `.enabled` property in favor of `.level`

### LINKED ISSUE #234: chalk.enabled=true or { enabled: true} does not work when color support is initially detected as false

Qix-:
Closes #234.

Also got to enable a test previously marking as failing, so that's cool. We were already pretty much entirely using `.level` so this was quite low-touch.


readme.md
```diff
@@ -120,21 +120,11 @@ Chain [styles](#styles) and call the last one as a method with a string argument
...
-
-Color support is automatically detected, as is the level (see `chalk.level`). However, if you'd like to simply enable/disable Chalk, you can do so via the `.enabled` property. When `chalk.enabled` is `true`, `chalk.level` must *also* be greater than `0` for colored output to be produced.
-
-Chalk is enabled by default unless explicitly disabled via `new chalk.Instance()` or `chalk.level` is `0`.
-
-If you need to change this in a reusable module, create a new instance:
-
-```js
-const ctx = new chalk.Instance({enabled: false});
-```
```
sindresorhus: I think we should preserve line 129 to 133 here with `level` instead.

Qix-: https://github.com/chalk/chalk/pull/356/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9L139

sindresorhus: Oh, I missed that since it was collapsed.

PR #367, opened 2019-09-30
Replace level list with table in readme.md

pedrottimark:
First of all, thank you for this excellent package <3

To use 256 background colors when available, I tripped over 0-based versus 1-based `level`

Although `readme.md` has 0-based numbered list, here is picture of the **list** on `npm` page:

chalk-level-list

Here is picture of the proposed **table** in markdown preview of code editor:

chalk-level-table

Updated picture according to **review** comments:

chalk-level-table-2

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -132,12 +132,12 @@ If you need to change this in a reusable module, create a new instance:
 const ctx = new chalk.Instance({level: 0});
 ```
 
-Levels are as follows:
-
-0. All colors disabled
-1. Basic color support (16 colors)
-2. 256 color support
-3. Truecolor support (16 million colors)
+| level | description |
+| ---: | :--- |
+| `0` | All colors disabled |
+| `1` | Basic color support (16 colors) |
+| `2` | 256 color support |
+| `3` | Truecolor support (16 million colors) |
 
 ### chalk.supportsColor
 
--- a/readme.md
+++ b/readme.md
@@ -132,8 +132,8 @@ If you need to change this in a reusable module, create a new instance:
 const ctx = new chalk.Instance({level: 0});
 ```
 
-| level | description |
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -132,12 +132,12 @@ If you need to change this in a reusable module, create a new instance:
...
 
-Levels are as follows:
-
-0. All colors disabled
-1. Basic color support (16 colors)
-2. 256 color support
-3. Truecolor support (16 million colors)
+| level | description |
+| ---: | :--- |
+| `0` | All colors disabled |
```
kevva: Could you center the numbers in the level column?

readme.md
```diff
@@ -132,12 +132,12 @@ If you need to change this in a reusable module, create a new instance:
 const ctx = new chalk.Instance({level: 0});
 ```
 
-Levels are as follows:
-
-0. All colors disabled
-1. Basic color support (16 colors)
-2. 256 color support
-3. Truecolor support (16 million colors)
+| level | description |
```
kevva: Use capitalization. `Level` and `Description`.

readme.md
```diff
@@ -132,12 +132,12 @@ If you need to change this in a reusable module, create a new instance:
 const ctx = new chalk.Instance({level: 0});
 ```
 
-Levels are as follows:
```
kevva: Why did you remove this?

pedrottimark: It seemed like the the heading row of the table communicates the same information.

I am happy to put it back if you prefer.

kevva: Indeed, I guess they do. I don't have a strong opinion about it so let's remove it for now unless somebody else wants to keep it.

Qix-:
Good catch, thank you :)

PR #368, opened 2019-10-07
Add `ansi256` and `bgAnsi256` to TypeScript declarations

pedrottimark:
When I test `bgAnsi256` as visual enhancement,  `2.4.2` and `3.0.0-beta.1` have TypeScript errors:

```text
ansi256.ts:3:30 - error TS2339: Property 'bgAnsi256' does not exist on type 'Chalk'.

3 const aColor = chalk.magenta.bgAnsi256(225);
                               ~~~~~~~~~

ansi256.ts:4:28 - error TS2339: Property 'bgAnsi256' does not exist on type 'Chalk'.

4 const bColor = chalk.green.bgAnsi256(194);
```

1. Add 2 type tests to `index.test-d.ts` which failed first
2. Add 2 declarations to `index.d.ts`
3. Add link and example to `readme.md`
4. Delete extra space in neighbor line of `readme.md`

PR #369, opened 2019-10-08
Add `ansi` and `bgAnsi` to TypeScript declaration

pedrottimark:
Here are remaining methods, after I researched the meaning of the argument:

1. Add 2 type tests to `index.test-d.ts` which failed first
2. Add 2 declarations to `index.d.ts`
3. Add link and example to `readme.md`

Thank you very much for quick turn-around on #368 ❤️

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -214,6 +214,14 @@ declare namespace chalk {
 		hwb(hue: number, whiteness: number, blackness: number): Chalk;
 
 		/**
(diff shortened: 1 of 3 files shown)


pedrottimark:
What do y’all think about removing these 2 methods from `chalk` public interface?

Through research and discussion in #370 I see internal use of `ansi` to convert down to lower-level format, but it seems like SGR color code as argument is a limited use case for `chalk` callers:

* first I jumped to assumption that argument is 4-bit unsigned color index, like `1` for red
* then I jumped to assumption that at level 1 the argument could be any SGR code, like `1` for bold, but the close code does not match: default color not normal weight, or whatever else

sindresorhus:
@pedrottimark Can you move the above into a new issue? The Chalk v3 release is imminent, but we can discuss whether to remove them for the next major version.

PR #375, opened 2019-11-12
Update dependents to ~48k

tlystad24:


Qix-:
I don't really think a 2k increase warrants a commit. It's only been a month since the last update. 🙃

sindresorhus:
No thanks

PR #379, opened 2019-12-10
Add link to replit run

eankeen:
i think it would be pretty cool to let people see how this code works (and edit / preview the library) right in their browser, without any manual downloads or installation. i'm thinking we can add a 'run on repl.it' button, which redirects to something like [this](https://repl.it/@eankeen/run-chalk):

image

i suppose if this is what we want, then we could possibly want the url to be owned by a more agnostic entity

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "nodejs"
+run = "cd examples && node rainbow"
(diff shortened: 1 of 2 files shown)


Qix-:
Deferring decision to @sindresorhus.

sindresorhus:
Sure :)

PR #385, opened 2020-02-24
Remove const from LevelEnum

AndrewLeedham:
`const` is not required in front of `enum`. TS does not let you reassign an `enum` anyway, so this should provide better compatibility.

fixes: #382

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-declare const enum LevelEnum {
+declare enum LevelEnum {
 	/**
 	All colors disabled.
 	*/


sindresorhus:
See https://github.com/chalk/chalk/pull/373

sindresorhus:
Duplicate of https://github.com/chalk/chalk/pull/373

sindresorhus:
> const is not required in front of enum. TS does not let you reassign an enum anyway, so this should provide better compatibility.

I don't think you understand what the `const` modifier does in this context: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

AndrewLeedham:
Fair point, did not come across that in my admittedly brief research.

PR #389, opened 2020-03-14
Add back apply, bind, and call from Function

### LINKED ISSUE #388: Missing standard Function prototype (e.g. chalk.apply is not a function)
I was logging a lot of things to stderr in my TypeScript project, so I wanted to create a neat wrapper for this:

```ts
function log (text: TemplateStringsArray, ...placeholders: unknown[]): void {
  console.error(chalk.stderr(text, ...placeholders))
}
```

Unfortunately, this yielded the following error:

```text
TypeError: chalk.apply is not a function
    at log (/foo/bar.ts:2:22)
```

I hadn't yet setup a `tsconfig`, so my project was targeting ES5. This meant that the `...placeholders` call was converted to use `.apply`, which is assumed to exist on every function.

However, since chalk changes the prototype of the function, there is no longer any `apply`, `call` or other things from the `Function` prototype.

I think that this could be easily resolved by adding in `apply`, `call`, etc. manually. But I wanted to open this to get some feedback first.

LinusU:
Fixes #388

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -41,6 +41,10 @@ const chalkFactory = options => {
 	Object.setPrototypeOf(chalk, Chalk.prototype);
 	Object.setPrototypeOf(chalk.template, chalk);
 
+	chalk.template.apply = Function.prototype.apply;
+	chalk.template.bind = Function.prototype.bind;
+	chalk.template.call = Function.prototype.call;
+
 	chalk.template.constructor = () => {
 		throw new Error('`chalk.constructor()` is deprecated. Use `new chalk.Instance()` instead.');
 	};
--- a/test/chalk.js
+++ b/test/chalk.js
@@ -110,3 +110,15 @@ test('sets correct level for chalk.stderr and respects it', t => {
 	t.is(chalk.stderr.level, 3);
 	t.is(chalk.stderr.red.bold('foo'), '\u001B[31m\u001B[1mfoo\u001B[22m\u001B[39m');
 });
+
+test('keeps .apply from Function', t => {
+	t.is(chalk.apply(chalk, ['foo']), 'foo');
+});
+
+test('keeps .bind from Function', t => {
+	t.is(chalk.bind(chalk, 'foo')(), 'foo');
+});
+
+test('keeps .call from Function', t => {
+	t.is(chalk.call(chalk, 'foo'), 'foo');
+});


Qix-:
I'm not at a machine right now that I can test this; does this apply to e.g. `chalk.red` or only the root `chalk` object?

ExE-Boss:
A better option might be to do:
```js
Object.setPrototypeOf(Chalk.prototype, Function.prototype);
```

LinusU:
@ExE-Boss Hmm, actually, `Chalk` is already a function, so it should already have the function prototype, but then why isn't it working as it is? 🤔 

```
function Chalk () { ... }

const chalk = {}
chalk.template = (...arguments_) => chalkTag(chalk.template, ...arguments_)

Object.setPrototypeOf(chalk, Chalk.prototype)
Object.setPrototypeOf(chalk.template, chalk)
```

Shouldn't the prototype chain here be `chalk.template` -> `chalk` -> `Chalk.prototype` -> `Function.prototype`? Or am I missing something? 🤔

ExE-Boss:
The prototype chain is: `chalk.template` → `chalk` → `Chalk.prototype` → `Object.prototype` → `null`.

LinusU:
Screenshot 2020-04-03 at 14 01 02

Hmm, I just don't understand why `Chalk.prototype` is pointing to `Object.prototype` when it's just defined as a function in the code. I mean obviously I'm missing something 😁

ExE-Boss:
That’s standard **ECMAScript** behaviour: <https://tc39.es/ecma262/#sec-makeconstructor>.

The `prototype` property of constructable functions is initialised to a plain object with its prototype set to `Object.prototype`.

This object is used [to create new instances by the `new` operator](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/new#Description) and [to check for inheritance by the `instanceof` operator](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/instanceof#Description).

LinusU:
Ahh, of course 🤦‍♂

Thanks for pointing me in the right direction!

Yeah, doing `Object.setPrototypeOf(Chalk.prototype, Function.prototype)` is probably a better option, will update the PR 👍

drumenbassik:
1

Qix-:
Ping @LinusU

sindresorhus:
Closing for lack of response.

LinusU:
Sorry for the lack of response. I don't personally need this anymore since I don't use `apply` now that `...` is natively supported.

PR #390, opened 2020-03-23
fix: correct type declarations in isolated modules mode

SimenB:
When compiling with `isolatedModules`, I get a type error from Chalk.

image

This change fixes it.

https://www.typescriptlang.org/v2/en/tsconfig#isolatedModules

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-declare const enum LevelEnum {
+declare enum LevelEnum {
 	/**
 	All colors disabled.
 	*/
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,4 +1,4 @@
-declare const enum LevelEnum {
```
SimenB: from my understanding `const enum` is a runtime optimization, and it should make no difference to the type declaration to remove it

package.json
```diff
@@ -60,5 +60,10 @@
 			"unicorn/prefer-string-slice": "off",
 			"unicorn/prefer-includes": "off"
 		}
+	},
+	"tsd": {
+		"compilerOptions": {
+			"isolatedModules": true
```
SimenB: by adding this, it fails without the change in this PR

image

Qix-:
Duplicate of #373

PR #380, opened 2019-12-22
Disallow template literals on colors and modifiers in the TypeScript types

sexwithsatan:
Fixes #341

PR #392, opened 2020-03-30
Support template literals for nested calls

### LINKED ISSUE #341: Template literals are unsupported for nested calls
[Issuehunt badges](https://issuehunt.io/r/chalk/chalk/issues/341)

The following is unsupported despite it being allowed by the type system:
```ts
chalk.bold`Hello, {cyan.inverse ${name}!} This is a test. {green ${exclamation}!}`;
```
Produces:
```
Hello, {cyan.inverse ,!} This is a test. {green ,!} Sindre Neat
```
Instead of:
```
Hello, Sindre! This is a test. Neat!
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Submitted pull Requests
- [#392 Support for template literals for nested calls](https://issuehunt.io/r/chalk/chalk/pull/392)
- [#398 Support nested calls](https://issuehunt.io/r/chalk/chalk/pull/398)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/chalk/chalk/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

toonijn:
Fixes #341
Closes #398 
Closes #380

Example:
```
chalk.red`2 + 3 = {bold ${2+3}}`
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#341: Template literals are unsupported for nested calls](https://issuehunt.io/repos/11855195/issues/341)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

### DIFF
--- a/benchmark.js
+++ b/benchmark.js
@@ -47,4 +47,16 @@ suite('chalk', () => {
 	bench('cached: 1 style nested non-intersecting', () => {
 		chalkBgRed(blueStyledString);
 	});
+
+	set('iterations', 10000);
+
+	bench('cached: 1 style template literal', () => {
+		// eslint-disable-next-line no-unused-expressions
+		chalkRed`the fox jumps over the lazy dog`;
+	});
+
+	bench('cached: nested styles template literal', () => {
+		// eslint-disable-next-line no-unused-expressions
+		chalkRed`the fox {bold jumps} over the {underline lazy} dog`;
+	});
 });
--- a/source/index.js
+++ b/source/index.js
@@ -6,6 +6,8 @@ const {
 	stringEncaseCRLFWithFirstIndex
 } = require('./util');
 
+const {isArray} = Array;
+
 // `supportsColor.level` → `ansiStyles.color[name]` mapping
 const levelMapping = [
 	'ansi',
@@ -134,7 +136,7 @@ const createStyler = (open, close, parent) => {
 
 const createBuilder = (self, _styler, _isEmpty) => {
 	const builder = (...arguments_) => {
-		if (Array.isArray(arguments_[0])) {
+		if (isArray(arguments_[0]) && isArray(arguments_[0].raw)) {
 			// Called as a template literal, for example: chalk.red`2 + 3 = {bold ${2+3}}`
 			return applyStyle(builder, chalkTag(builder, ...arguments_));
 		}
@@ -193,7 +195,7 @@ let template;
 const chalkTag = (chalk, ...strings) => {
 	const [firstString] = strings;
 
-	if (!Array.isArray(firstString)) {
+	if (!isArray(firstString) || !isArray(firstString.raw)) {
 		// If chalk() was called by itself or with a string,
 		// return the string itself as a string.
 		return strings.join(' ');
--- a/source/index.js
+++ b/source/index.js
@@ -135,7 +135,7 @@ const createStyler = (open, close, parent) => {
 const createBuilder = (self, _styler, _isEmpty) => {
 	const builder = (...arguments_) => {
 		if (Array.isArray(arguments_[0])) {
-			// Called as a template litteral, e.g. chalk.red`2 + 3 = {bold ${2+3}}`
+			// Called as a template literal, e.g. chalk.red`2 + 3 = {bold ${2+3}}`
 			return applyStyle(builder, chalkTag(builder, ...arguments_));
 		}
 
--- a/source/index.js
+++ b/source/index.js
@@ -134,6 +134,11 @@ const createStyler = (open, close, parent) => {
 
 const createBuilder = (self, _styler, _isEmpty) => {
 	const builder = (...arguments_) => {
+		if (Array.isArray(arguments_[0])) {
+			// Called as a template litteral, e.g. chalk.red`2 + 3 = {bold ${2+3}}`
+			return applyStyle(builder, chalkTag(builder, ...arguments_));
+		}
+
 		// Single argument is hot path, implicit coercion is faster than anything
 		// eslint-disable-next-line no-implicit-coercion
 		return applyStyle(builder, (arguments_.length === 1) ? ('' + arguments_[0]) : arguments_.join(' '));
--- a/source/index.js
+++ b/source/index.js
@@ -135,7 +135,7 @@ const createStyler = (open, close, parent) => {
 const createBuilder = (self, _styler, _isEmpty) => {
 	const builder = (...arguments_) => {
 		if (Array.isArray(arguments_[0])) {
-			// Called as a template literal, e.g. chalk.red`2 + 3 = {bold ${2+3}}`
+			// Called as a template literal, for example: chalk.red`2 + 3 = {bold ${2+3}}`
 			return applyStyle(builder, chalkTag(builder, ...arguments_));
 		}
 
--- a/test/template-literal.js
+++ b/test/template-literal.js
@@ -42,6 +42,11 @@ test('correctly perform nested template substitutions', t => {
 
 	t.is(instance.strikethrough.cyanBright.bgBlack`Works with {reset {bold numbers}} {bold.red ${1}}`,
 		instance.strikethrough.cyanBright.bgBlack('Works with ' + instance.reset.bold('numbers') + ' ' + instance.bold.red(1)));
+
+	t.is(chalk.bold`Also works on the shared {bgBlue chalk} object`,
+		'\u001B[1mAlso works on the shared \u001B[1m' +
+		'\u001B[44mchalk\u001B[49m\u001B[22m' +
+		'\u001B[1m object\u001B[22m');
 });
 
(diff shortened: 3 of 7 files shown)


sindresorhus:
I don’t see any mention of performance here. Please see the concerns in the issue.

toonijn:
You're right, sorry.

In the usual code execution only one if-clause extra is checked:
```
if(Array.isArray(arguments_[0])) {
    ... // the new code
}
```

This clause is only triggered when the function (e.g. `chalk.red`) is called with an array argument. This (only) happens when used as a template literal. So the speed of existing code is not impacted.

Qix-:
`if` branches are **not** free. Please do a proper performance benchmark.

toonijn:
Using 'npm run bench'. These are the results

```
(nested_template_literals)
      24,755,209 op/s » 1 style
      21,323,845 op/s » 2 styles
      20,138,529 op/s » 3 styles
      23,569,342 op/s » cached: 1 style
      22,033,244 op/s » cached: 2 styles
      19,742,892 op/s » cached: 3 styles
      11,052,769 op/s » cached: 1 style with newline
       5,337,312 op/s » cached: 1 style nested intersecting
      14,257,177 op/s » cached: 1 style nested non-intersecting


```
(master)
      25,268,827 op/s » 1 style
      23,094,279 op/s » 2 styles
      20,585,710 op/s » 3 styles
      25,472,874 op/s » cached: 1 style
      22,120,499 op/s » cached: 2 styles
      20,772,105 op/s » cached: 3 styles
      11,559,382 op/s » cached: 1 style with newline
       5,340,114 op/s » cached: 1 style nested intersecting
      14,327,385 op/s » cached: 1 style nested non-intersecting
```

**Comment by toonijn** [comment]
Are there still issues with this PR?

File: source/index.js [hunk 1]
```diff
@@ -134,6 +134,11 @@ const createStyler = (open, close, parent) => {
 
 const createBuilder = (self, _styler, _isEmpty) => {
 	const builder = (...arguments_) => {
+		if (Array.isArray(arguments_[0])) {
+			// Called as a template litteral, e.g. chalk.red`2 + 3 = {bold ${2+3}}`
```
sindresorhus, line 138: typo

test/template-literal.js
```diff
@@ -30,6 +30,14 @@ test('correctly perform template substitutions', t => {
 		instance.bold('Hello,', instance.cyan.inverse(name + '!'), 'This is a') + ' test. ' + instance.green(exclamation + '!'));
 });
 
+test('correctly perform nested template substitutions', t => {
+	const instance = new chalk.Instance({level: 0});
+	const name = 'Sindre';
+	const exclamation = 'Neat';
+	t.is(instance.bold`Hello, {cyan.inverse ${name}!} This is a` + ' test. ' + instance.green`${exclamation}!`,
+		instance.bold('Hello,', instance.cyan.inverse(name + '!'), 'This is a') + ' test. ' + instance.green(exclamation + '!'));
+});
```
sindresorhus, line 39: Can you also do a test that uses more nested properties? Like:

```
chalk.bold.red.underline`Hello`;
```

sindresorhus:
Can you add some tests to the benchmark using template literals and nesting? Could be useful for the future.

sindresorhus:
This also needs to be documented in the readme and index.d.ts.

toonijn:
- index.d.ts already allowed template literals for chained styles. This is now also reflected in index.test-d.ts.
- The readme contains an explicit example of the use of `` chalk.bold.rgb(10, 100, 200)`Hello!` ``
- Some benchmarks are added for template literals. Parsing templates is quite expensive so I've lowered the number of iterations for these two benchmarks
- Two complex chained template literals are tested.

sindresorhus:
> index.d.ts already allowed template literals for chained styles.

I meant docs-wise. The readme and index.d.ts should be in sync docs-wise.

toonijn:
Is it okay like this?

sindresorhus:
@toonijn I don't understand your question. I already commented what needs to be done.

toonijn:
I think I've implemented the changes in my latest commit.
What docs should I change more?

sindresorhus:
I already commented that. What you added to the readme, you need to add to index.d.ts too (docs-wise).

toonijn:
@sindresorhus A small example is included in index.d.ts as well.

test/template-literal.js
```diff
@@ -30,6 +30,20 @@ test('correctly perform template substitutions', t => {
...
+	const exclamation = 'Neat';
+	t.is(instance.bold`Hello, {cyan.inverse ${name}!} This is a` + ' test. ' + instance.green`${exclamation}!`,
+		instance.bold('Hello,', instance.cyan.inverse(name + '!'), 'This is a') + ' test. ' + instance.green(exclamation + '!'));
+
+	t.is(instance.red.bgGreen.bold`Hello {italic.blue ${name}}`,
+		instance.red.bgGreen.bold('Hello ' + instance.italic.blue(name)));
+
+	t.is(instance.strikethrough.cyanBright.bgBlack`Works with {reset {bold numbers}} {bold.red ${1}}`,
+		instance.strikethrough.cyanBright.bgBlack('Works with ' + instance.reset.bold('numbers') + ' ' + instance.bold.red(1)));
+});
```
sindresorhus, line 45: Can you also add a separate test that uses the shared Chalk instance not `new chalk.Instance`?

sindresorhus:
I'm ok with this.

test/template-literal.js
Qix-: @toonijn this is marked resolved but I don't see where this is happening.

source/index.js
```diff
@@ -134,6 +134,11 @@ const createStyler = (open, close, parent) => {
 
 const createBuilder = (self, _styler, _isEmpty) => {
 	const builder = (...arguments_) => {
+		if (Array.isArray(arguments_[0])) {
```
Qix-, line 137:
```suggestion
		if (Array.isArray(arguments_[0]) && arguments_[0].raw) {
```

benchmark.js
```diff
@@ -47,4 +47,16 @@ suite('chalk', () => {
 	bench('cached: 1 style nested non-intersecting', () => {
 		chalkBgRed(blueStyledString);
 	});
+
+	set('iterations', 10000);
```
Qix-, line 51: Why manually set iterations here? This can skew results if too low.

source/index.js
Qix-: Otherwise, this would break any sort of usecase where people are relying on array -> string coercion.

sindresorhus: ^ Would be good to have a test for this too.

ExE-Boss: It migh also be a good idea to cache `Array.isArray` in a module scope `const` to minimise `[[Get]]` lookups of a mutable property:
```suggestion
		if (ArrayIsArray(arguments_[0]) && ArrayIsArray(arguments_[0].raw)) {
```

test/template-literal.js
toonijn: It wasn't resolved yet. Now it should be.

source/index.js
toonijn: - I added a small test to see if chalk correctly casts int and array to string. This revealed a previously existing bug: `chalk(["abc"])` threw an error. This should now be fixed by adding the suggestion on line 198  as well https://github.com/chalk/chalk/blob/26c73d2b80f06f5673a7bb44433f69316f229629/source/index.js#L198
- Caching Array.isArray is indeed faster. (from 24M ops/s to 25M ops/s for the first benchmark).

benchmark.js
toonijn: Parsing strings for the special `{bold ...}`-syntax is a lot slower.
Benchmark 'cached: 1 style' is almost 200 times faster than 'cached: 1 style template literal'. So I reduced the number of iterations accordingly.

Better than specifying the number of iterations would be to set a minimal runtime for each benchmark. So instead of
https://github.com/chalk/chalk/blob/26c73d2b80f06f5673a7bb44433f69316f229629/benchmark.js#L6
something like
```
set('mintime', 500); 
```

But this feels out of the scope of this PR.

Qix-: > Better than specifying the number of iterations would be to set a minimal runtime for each benchmark.

I agree. Mind opening a PR? :)

test/template-literal.js
Qix-: No worries; I'm not used to people properly using the Resolve feature on Github :D

sindresorhus:
Thanks :)

PR #393, opened 2020-03-31
Properly check if `level` is an integer

Richienb:


Qix-:
Why is this necessary?

sindresorhus:
@Qix- To make it as strict as possible. Yes, people do pass floats to APIs like this. Don't ask me why.

PR #396, opened 2020-04-15
Update readme links to use https

styfle:
This should fix the "Not Secure" indicator on Google Chrome address bar.

image


Qix-:
Thanks @styfle :)

PR #414, opened 2020-10-03
Update test dependencies

Richienb:


Qix-:
I don't agree with a number of these changes, personally - namely some of the changes in import semantics and the syntactic changes of arrow functions. I'll let @sindresorhus decide though, because they may be related to Ava and/or XO, both of which he maintains.

Richienb:
@Qix- The import statements become require statements because ava 3.x no longer supports esm and the syntactic changes of arrow functions are enforced by the newest version of xo.

Qix-:
Interesting, I don't use Ava myself so I wasn't aware. Even more reason for @sindresorhus to review, then :)

sindresorhus:
I'm holding off on upgrading AVA until next April when I can just convert everything to native ESM. If we convert all tests to require now, it will just be a lot of churn when we convert back again in half a year. I'll take the other changes though.

package.json
```diff
@@ -62,7 +62,8 @@
 			"@typescript-eslint/member-ordering": "off",
 			"no-redeclare": "off",
 			"unicorn/string-content": "off",
-			"unicorn/better-regex": "off"
+			"unicorn/better-regex": "off",
+			"unicorn/prefer-number-properties": "off"
```
sindresorhus, line 66: ?

Richienb: In `templates.js`, can we replace `parseInt` with `Number.parseInt`?

sindresorhus: Yes

Richienb: ok

Qix-:
Thanks @Richienb :)

PR #420, opened 2020-11-19
Adds travis jobs on ppc64le

dineshks1:
This PR adds power support on travis.


Qix-:
We've already denied this before. No thanks.

Issue #431, opened 2021-04-16
Chalk v5


sindresorhus:
- [x] ESM (https://github.com/chalk/chalk/commit/fa16f4ec37c1c25bda6ca1864bf53ecd3ea3d3c9)
- [x] Remove some color types, per https://github.com/chalk/ansi-styles/commit/9150f611ced8afc843741050dfb05f48aad1d0dc
- [x] Make `chalk.stderr` a named export instead? `import {chalkStderr} from 'chalk';`
- [x] Make `chalk.supportsColor` a named export instead? `import {supportsColor} from 'chalk';`.
- [x] Make `chalk.Instance` a named export instead? `import {Chalk} from 'chalk';`.
- [x] Move template functionality to [`chalk-template`](https://github.com/chalk/chalk-template).
- [x] Clean up the type definition. Improve the type names. Add more docs.

What else do we want to do for v5?

---

/// @Qix- @Richienb

Richienb:
TypeScript rewrite would be nice

Qix-:
This all sounds good to me. There isn't anything else pressing that I can think of.

I'm :-1: on the typescript rewrite. I don't see what benefits that would bring.

Richienb:
All the benefits of rewriting in TypeScript are maintainer facing.

It allows us to automatically type check our code and generate typings for distribution.

sindresorhus:
I don't think we want TypeScript for Chalk. Chalk is just doing too much weird stuff that is hard to type. It would also slightly bloat Chalk installation-wise.

Richienb:
Would it make sense to use symbols instead of underscored properties?

https://github.com/chalk/chalk/blob/d798222a5a3ff208782ce81f5fbd3babb286b3e3/source/index.js#L149-L151

sindresorhus:
Yes, I think that's a good idea.

But if you want to contribute more, "Remove some color types, per chalk/ansi-styles@9150f61" is a higher priority.

fregante:
This? https://github.com/chalk/chalk/issues/300#issuecomment-730972576

Qix-:
As much as it pains me to suggest this since it's my baby...

Perhaps we move the template code into `chalk-template` or something similar (if @smollweide would allow for that, since the existing package seems to have gone untouched for a while). While I still assert it's incredibly useful functionality, I haven't seen a lot of people use it and it would reduce install times and extra init costs even more.

[...]

Issue #370, opened 2019-10-14
To upsample, or not to upsample?


pedrottimark:
Please read this as an observation from a thankful user, instead of a late request for chalk 3

It might be relevant input to your discussion about performance and simplification

> Colors are downsampled from 16 million RGB values to an ANSI color format that is supported by the terminal emulator

While studying `bgAnsi256` as a visual enhancement when available, I noticed that `chalk` also **upsamples** from lower-level color models to higher-level ANSI color formats

It looks like a side-effect of the implementation more than an intended feature, because:

* less predictable result (see **ansi** below)
* more complicated to test
* obstacle to faster paths
* more than necessary dependence on `color-convert`

### green

Here is `(new chalk.Instance({level)).green('…')` with `^` as placeholder for escape:

| level | output |
| ---: | :--- |
| `0` | `…` |
| `1` | `^[32m…^[39m` |
| `2` | `^[32m…^[39m` |
| `3` | `^[32m…^[39m` |

This is predictable: it uses lowest-level format and downsamples when needed

### ansi

Here is `(new chalk.Instance({level)).ansi(32)('…')` which is a synonym for green:

| level | output |
| ---: | :--- |
| `0` | `…` |
| `1` | `^[32m…^[39m` |
| `2` | `^[38;5;34m…^[39m` |
| `3` | `^[38;2;0;128;0m…^[39m` |

The visual result is less predictable because on macOS 10.12.6 Terminal:

* level `1` output has color `#17a41a` from default theme
* level `2` output has color `#19ad1d`

This is research, because I don’t need to use `ansi` nor `bgAnsi`

### ansi256

Here is `(new chalk.Instance({level)).bgAnsi256(194)('…')`

| level | output |
| ---: | :--- |
| `0` | `…` |
| `1` | `^[107m…^[49m` |
| `2` | `^[48;5;194m…^[49m` |
| `3` | `^[48;2;204;255;204m…^[49m` |

Because the contrasting background colors that I intend to use both downsample to `107` I would use `chalk.level >= 2` as condition for foreground color with or without background color

However, it would eliminate a test case if levels `2` and `3` have the same format

pedrottimark:
Here are the relevant code snippets from the `index.js` file:

```js
const levelMapping = [
  'ansi',
  'ansi',
  'ansi256',
  'ansi16m'
];
```

Here are places for potential short-cut to avoid upsample for some `level` and `model`:

[...]

PR #433, opened 2021-04-17
Add `overline` style and remove `keyword`, `hsl`, `hsv`, `hwb` and `ansi` color spaces

### LINKED ISSUE #360: Remove some color types
HCG was added after `color-convert` was added to ansi-styles. We should officially support it in `ansi-styles` and document its support here. It's similar to HSL and is (slowly) increasing in popularity.

Richienb:
# Update `ansi-styles` package
- Add `overline` style
- Remove `keyword`, `hsl`, `hsv`, `hwb` and `ansi` color spaces

#431 

Fixes #360

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -217,6 +217,7 @@ export interface ChalkInstance extends ChalkFunction {
 
 	/**
 	Use a [Select/Set Graphic Rendition](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters) (SGR) [color code number](https://en.wikipedia.org/wiki/ANSI_escape_code#3/4_bit) to set background color.
+
 	30 <= code && code < 38 || 90 <= code && code < 98
 	For example, 31 for red, 91 for redBright.
 	Use the foreground code, not the background code (for example, not 41, nor 101).
--- a/index.d.ts
+++ b/index.d.ts
@@ -184,7 +184,7 @@ export interface ChalkInstance extends ChalkFunction {
 	hex: (color: string) => this;
 
 	/**
-	Use a [8-bit unsigned number](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) to set text color.
+	Use an [8-bit unsigned number](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) to set text color.
 	*/
 	ansi256: (index: number) => this;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -61,6 +61,7 @@ export type Modifiers =
 	| 'dim'
 	| 'italic'
 	| 'underline'
+	| 'overline'
 	| 'inverse'
 	| 'hidden'
 	| 'strikethrough'
@@ -164,6 +165,11 @@ export interface ChalkInstance extends ChalkFunction {
 	level: ColorSupportLevel;
 
 	/**
+	Use RGB values to set text color.
+	*/
+	rgb: (red: number, green: number, blue: number) => this;
+
+	/**
 	Use HEX value to set text color.
 
 	@param color - Hexadecimal value representing the desired color.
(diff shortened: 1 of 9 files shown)


Richienb:
Because desampling to `ansi` is no longer active, level 1 is now effectively the same in function as level 2.

Qix-:
Just curious, why aren't we downsampling anymore? That seems like a functional regression, given that there are still _loads_ of terminals that only support the older 16 color palette.

Richienb:
Because you said so

Qix-:
I don't see in there why we need to remove this functionality. I would say we should inline that from `color-convert` (you're free to copy over the `ansi256.ansi16` function from `conversions.js` directly - not that the license would have prevented you anyway). Just mention it came from `color-convert` in a comment.

The concern about upscaling (#370) was valid, but downscaling is absolutely necessary to keep compatibility with people who do not have 256 support.

Richienb:
> you're free to copy over the `ansi256.ansi16` function from `conversions.js` directly - not that the license would have prevented you anyway

That function doesn't exist

Qix-:
Right, you have to merge them - this is untested but should be close enough. Just needs a bit of polish for the Chalk conventions.

```javascript
function ansi256To16(args) {
	let r, g, b;

	// Handle greyscale
	if (args >= 232) {
		const c = (args - 232) * 10 + 8;
		r = g = b = c / 255;
	} else {
		args -= 16;

		let rem;
		r = Math.floor(args / 36) / 5;
		g = Math.floor((rem = args % 36) / 6) / 5;
		b = (rem % 6) / 5;
	}

        const value = Math.max(r, g, b) * 2;

	if (value === 0) {
		return 30;
	}

	let ansi = 30
		+ ((Math.round(b) << 2)
		| (Math.round(g) << 1)
		| Math.round(r));

	if (value === 2) {
		ansi += 60;
	}

	return ansi;
}
```

Richienb:
Support for `ansi` has been added back in.

Qix- (approved):
I don't see anything glaringly wrong with the code; I'll let Sindre have the final look.

sindresorhus:
I'm confused. I understand preserving the downsampling, but shouldn't we still remove the explicit `.ansi()` method?

Richienb:
IMO, there is no impact of leaving the function there so we might as well leave it for people wanting to use the `ansi` color space.

Qix-:
People wanting to use the `ansi` color space should just use `.red()`, `.blueBright()`, etc as it's a 1:1 mapping. People doing programmatic generation of colors for whatever reason are going to be using 256 or truecolor - not 16 - as there is no logical rhyme or reason to the code (I always remember them as "traffic light, dark blue, pink, light blue, white" - completely arbitrary).

I agree with sindre here - no need for `.ansi()`. Chalk is a utility library, after all - you're always able to write escapes yourself with trivial syntax.

Richienb:
Removed

sindresorhus (approved):

index.d.ts
```diff
@@ -178,51 +184,14 @@ export interface ChalkInstance extends ChalkFunction {
...
-	Use HWB values to set text color.
-	*/
-	hwb: (hue: number, whiteness: number, blackness: number) => this;
-
-	/**
-	Use a [Select/Set Graphic Rendition](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters) (SGR) [color code number](https://en.wikipedia.org/wiki/ANSI_escape_code#3/4_bit) to set text color.
-
-	30 <= code && code < 38 || 90 <= code && code < 98
-	For example, 31 for red, 91 for redBright.
+	Use a [8-bit unsigned number](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) to set text color.
```
Qix-, line 187:
```suggestion
	Use an [8-bit unsigned number](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) to set text color.
```

Qix- (changes requested):

Qix- (approved):
LGTM

sindresorhus:
Nice work, @Richienb 👍

Qix-:
Yes indeed :) Thanks @Richienb, super appreciated.

PR #437, opened 2021-04-22
Minor tweaks

Richienb:


source/index.js
```diff
@@ -9,6 +9,10 @@ import template from './templates.js';
 const {stdout: stdoutColor, stderr: stderrColor} = supportsColor;
 const {isArray} = Array;
 
+const GENERATOR = Symbol('Generator');
```
sindresorhus, line 12: I prefer to keep the symbol string the exact same as the variable.

sindresorhus (changes requested):

sindresorhus (approved):

index.d.ts
```diff
@@ -223,7 +223,7 @@ export interface ChalkInstance extends ChalkFunction {
 	readonly bold: this;
 
 	/**
-	Modifier: Emitting only a small amount of light.
+	Modifier: Make text lower opacity.
```
Qix-, line 226: It's not opacity, that's misleading. Also, it's not consistent across emulators, so we should note that. Perhaps something like...

```suggestion
	Modifier: Darken the text slightly. (Inconsistent across emulators; might do nothing)
```

Qix- (changes requested):
Just a nit, but otherwise this looks great.

Qix- (approved):

PR #499, opened 2021-07-23
docs: ✏️ add chinese readme

Muromi-Rikka:


sindresorhus:
Thanks for the pull request, but we don't have any plans to localize the docs.

PR #518, opened 2021-10-17
Fix Browser Colors

CEbbinghaus:
In b31d6fb48a570d9be48044abe656c36dcac094e4 it is mentioned that the browser has built-in support for ANSI color codes (which is true). But the reality is that using Chalk doesn't produce colors in the Devtools Console.

This was due to the `supports-color` package trying to identify the environment variables that node provides (which Browsers do not) therefore wrongfully assuming that there is no color support at all.

This change serves to allow for chalk to work in a web environment via Webpack or Snowpack.

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -6,7 +6,27 @@ import {
 } from './util.js';
 import template from './templates.js';
 
-const {stdout: stdoutColor, stderr: stderrColor} = supportsColor;
+// True for any Browser or WebWorker
+// copied from https://github.com/flexdinesh/browser-or-node/blob/master/src/index.js available under the MIT license
+const isBrowser =
+	(typeof window !== "undefined" && typeof window.document !== "undefined")
+	||
+	(typeof self === "object" &&
+	self.constructor &&
+	self.constructor.name === "DedicatedWorkerGlobalScope");
+
+// Overwrite SupportsColor if we are running in Browser or WebWorker
+const { stdout: stdoutColor, stderr: stderrColor } = !isBrowser
+	? supportsColor
+	: {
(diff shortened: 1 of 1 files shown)


sindresorhus:
This is already done in `supports-color`: https://github.com/chalk/supports-color/blob/main/browser.js

CEbbinghaus:
@sindresorhus 
The package.json for `supports-color` specifies that if it's included via node it only uses the index.js file and the only time it allows the browser.js file to be accessed is if the code is not running in a node environment. This falls over when a packaging tool such as Webpack is used however because it includes and builds the file server side before outputting the result in browser-compatible form so, in the end, the version that makes it into the final version is the basic index.html terminal check.

https://github.com/chalk/supports-color/blob/ff1704d46cfb0714003f53c8d7e55736d8d545ff/package.json#L14-L16

sindresorhus:
Then you haven't configured your Webpack config correctly. Webpack should pick the correct one when building for web. That's the whole point of having multiple exports in package.json.

CEbbinghaus:
> ven't configured your Webpack config correctly. Webpack should pick the correct one when building for web. That's the whole point of having multiple exports in package.json.

I'm using Typescript & Webpack so the first step is for typescript to compile the package and then webpack bundles that output

CEbbinghaus:
> Then you haven't configured your Webpack config correctly. Webpack should pick the correct one when building for web. That's the whole point of having multiple exports in package.json.

It seems that all it was was the `aliasFields: ["browser"],` option under the require object. I didn't realize that this was a valid solution. Thank you for the tip

PR #524, opened 2021-11-04
Remove support for tagged template literals

Richienb:
Related to https://github.com/chalk/chalk/issues/431


Qix- (approved):
I can't speak for the typescript changes but from what I can tell this looks right.

PR #526, opened 2021-11-21
Re-export types from `supports-color`

Richienb:


source/index.d.ts
```diff
@@ -67,14 +69,11 @@ export type Modifiers =
...
-- `1` - Basic 16 colors support.
-- `2` - ANSI 256 colors support.
-- `3` - Truecolor 16 million colors support.
-*/
-export type ColorSupportLevel = 0 | 1 | 2 | 3;
+export {
+	ColorInfo,
+	ColorSupport,
+	ColorSupportLevel,
+};
```
sindresorhus, line 76:
```suggestion
} from 'supports-color';
```

I prefer explicit re-exporting.

And place it at the bottom of the file.

PR #539, opened 2022-03-17
round r,g,b values in 16-million color support

Jakobeha:
Currently if you pass decimal values to rgb or bgRgb, it will generate output with those decimals, which iTerm silently ignores.

Not sure if this is the ideal fix, since when rgb values are from 0-255 it's *commonly* assumed that they're integers. Maybe chalk should just throw an error instead. But this is a bug I came across when converting from 0-1 r,g,b, which is common, so there should be at least some indication.


Qix-:
Hey, thanks for the PR. I'm not really interested in this change as this is going to severely slow down each individual call, and weighed against the fact this is an edge case, the tradeoff doesn't seem worth it.

The inputs must always be integers from 0..255. Maybe that isn't clear enough in the readme - a PR to improve that documentation is definitely welcome if you think we could do better there!

PR #564, opened 2022-09-23
Use template literal types for simple color name transformations.

m4ttsch:
For simple type transformations - `<color> -> bg<Color>`, for example - we can use [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) instead of multiple string literal type unions. An advantage of this approach is that the template literal types are always derived from the interpolated types, which means that we don't have to worry about missing types or misspelling types when performing these transformations.


Qix-:
I'm not against these I suppose. Didn't even know this was possible.

PR #566, opened 2022-10-03
Expose styles as variables

### LINKED ISSUE #565: Expose styles as variables?
A package I am maintaining [chalk-pipe](https://github.com/LitoMore/chalk-pipe) has a function to determine what type of style is passed in.

Currently Chalk only exposes the types of styles. But in my case, I want the variables.

https://github.com/LitoMore/chalk-pipe/blob/main/source/index.ts#L3-L23

Could we add some exports like this:

```js
export const modifiers = ['rest', 'bold' , 'dim', ...];
export const colors = = ['black', 'red', 'green', ...];
```

Since we already documented which styles are supported in the readme. It's a good idea to expose these styles as variables.

LitoMore:
Resolves #565

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -205,7 +205,7 @@ Explicit 256/Truecolor mode can be enabled using the `--color=256` and `--color=
 
 All supported style strings are exposed as an array of strings for convenience. `colors` is the combination of `foregroundColors` and `backgroundColors`.
 
-This is useful when checking if a style string is supported:
+This can be useful if you wrap Chalk and need to validate input:
 
 ```js
 import {modifiers, foregroundColors} from 'chalk';
--- a/readme.md
+++ b/readme.md
@@ -201,7 +201,7 @@ Explicit 256/Truecolor mode can be enabled using the `--color=256` and `--color=
 
 `chalkStderr` contains a separate instance configured with color support detected for `stderr` stream instead of `stdout`. Override rules from `supportsColor` apply to this too. `supportsColorStderr` is exposed for convenience.
 
-### modifiers, foregroundColors, backgroundColors and colors
+### modifiers, foregroundColors, backgroundColors, and colors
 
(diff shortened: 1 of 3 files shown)


source/index.js
```diff
@@ -209,4 +209,8 @@ export {
 	stderrColor as supportsColorStderr,
 };
 
+export const modifiers = Object.keys(ansiStyles.modifier);
+export const foregroundColors = Object.keys(ansiStyles.color);
+export const backgroundColors = Object.keys(ansiStyles.bgColor);
```
LitoMore, line 214: This makes our Chalk slower. Is this acceptable?

Qix-: That's very strange. Can you show how?

LitoMore: This makes the code slightly slower, but it might be negligible.

sindresorhus, line 214: Need to export `colors` too. Combination of foreground and background.

sindresorhus:
This needs to be documented.

source/index.d.ts
```diff
@@ -284,4 +284,8 @@ export {
 // } from '#supports-color';
 } from './vendor/supports-color/index.js';
 
+export const modifiers: Modifiers[];
```
sindresorhus, line 287:
```suggestion
export const modifiers: readonly Modifiers[];
```

readme.md
```diff
@@ -201,6 +201,10 @@ Explicit 256/Truecolor mode can be enabled using the `--color=256` and `--color=
 
 `chalkStderr` contains a separate instance configured with color support detected for `stderr` stream instead of `stdout`. Override rules from `supportsColor` apply to this too. `supportsColorStderr` is exposed for convenience.
 
+### modifiers, foregroundColors, backgroundColors, and colors
+
+All supported style strings are exposed for convenience. `colors` is the combination of `foregroundColors` and `backgroundColors`.
+
```
sindresorhus, line 207: I think an example could be useful, as it's not clear what these properties contain or what to use them for.

PR #567, opened 2022-10-05
Export styles from `ansi-styles`

LitoMore:
`ansi-styles` adds some extra keys to `styles`:

https://github.com/chalk/chalk/blob/92c55db46f2396c18764e55e6a52dcb49884a42b/source/vendor/ansi-styles/index.js#L98-L103

We should filter them out.

Or we could process those style names in `ansi-sytle`, see https://github.com/chalk/ansi-styles/pull/82. I marked that PR as a draft to discuss how we can improve on this.

My thoughts are:

~~**Option 1:**~~
~~Filter style names in Chalk~~

**Option 2:**
Apply https://github.com/chalk/ansi-styles/pull/82, then export style names from `ansi-styles` in Chalk

~~**Option 3:**~~
~~Apply https://github.com/chalk/ansi-styles/pull/82 change but in favor of modify the `vendor/ansi-styles` directly, then export styles names~~

What do you think?

For the exposed style names, we'd better add some tests to ensure its outputs are correct.

### DIFF
--- a/source/index.d.ts
+++ b/source/index.d.ts
@@ -242,10 +242,8 @@ export const chalkStderr: typeof chalk;
 export const supportsColorStderr: typeof supportsColor;
 
 export {
-	ModifierName,
-	ForegroundColorName,
-	BackgroundColorName,
-	ColorName,
+	ModifierName, ForegroundColorName, BackgroundColorName, ColorName,
+	modifierNames, foregroundColorNames, backgroundColorNames, colorNames,
 // } from '#ansi-styles';
 } from './vendor/ansi-styles/index.js';
 
@@ -255,26 +253,6 @@ export {
 	ColorSupportLevel,
 // } from '#supports-color';
 } from './vendor/supports-color/index.js';
-
-/**
-Basic modifier names.
-*/
-export const modifierNames: readonly ModifierName[];
-
-/**
-Basic foreground color names.
-*/
-export const foregroundColorNames: readonly ForegroundColorName[];
-
-/**
-Basic background color names.
-*/
-export const backgroundColorNames: readonly BackgroundColorName[];
-
-/**
-Basic color names. The combination of foreground and background color names.
-*/
-export const colorNames: readonly ColorName[];
 
 /**
 @deprecated Use `ModifierName` instead.
--- a/source/index.d.ts
+++ b/source/index.d.ts
@@ -1,7 +1,7 @@
 // TODO: Make it this when TS suports that.
-// import {ModifierName as Modifiers, ForegroundColorName as ForegroundColor, BackgroundColorName as BackgroundColor, ColorName as Color} from '#ansi-styles';
+// import {ModifierName, ForegroundColor, BackgroundColor, ColorName} from '#ansi-styles';
 // import {ColorInfo, ColorSupportLevel} from '#supports-color';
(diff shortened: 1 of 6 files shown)


sindresorhus:
I would go with option 2.

Option 3 is not entirely clear to me.

LitoMore:
> Option 3 is not entirely clear to me.

I mean change the code directly in https://github.com/chalk/chalk/blob/main/source/vendor/ansi-styles/index.js. But I don't think it's a good solution.

sindresorhus:
> I mean change the code directly in [https://github.com/chalk/chalk/blob/main/source/vendor/ansi-styles/index.js](https://github.com/chalk/chalk/blob/main/source/vendor/ansi-styles/index.js?rgh-link-date=2022-10-09T16%3A03%3A50Z). But I don't think it's a good solution.

No, we shouldn't do that. The vendor stuff is meant to be kept in sync with the source.

LitoMore:
Of course. I marked the https://github.com/chalk/ansi-styles/pull/82 as ready, please review.

I will sync those changes to this project.

sindresorhus:
Can you change the exports to use the `Name` suffix like `ansi-styles`? We already shipped the current names, so we need to preserve those using an alias with a `@deprecated` doc comment.

LitoMore:
@sindresorhus Done. This failing CI doesn't seem to be related to the code.

Qix-:
https://github.com/chalk/chalk/blob/main/.github/workflows/main.yml#L25

Not sure the goal here with this check but seems like codecov is completely broken regardless of version, it's just that it's only enabled on 16.

LitoMore:
Passing now.

Qix-:
Curious, what was the issue? The error message led me to believe it was a transient issue...

source/index.d.ts
```diff
@@ -277,16 +241,101 @@ export const supportsColor: ColorInfo;
 export const chalkStderr: typeof chalk;
 export const supportsColorStderr: typeof supportsColor;
 
+export {
+	ModifierName,
```
sindresorhus, line 245: You can use `export` for `modifierNames` and the other instead of redefining them and duplicating all the docs.

source/index.js
```diff
@@ -204,14 +204,20 @@ Object.defineProperties(createChalk.prototype, styles);
 const chalk = createChalk();
 export const chalkStderr = createChalk({level: stderrColor ? stderrColor.level : 0});
 
+export {
+	modifierNames,
+	foregroundColorNames,
+	backgroundColorNames,
+	colorNames,
+	modifierNames as modifiers,
```
sindresorhus, line 212: Add TODO comment to remove this in the next major version.

source/index.d.ts
```diff
@@ -277,16 +241,101 @@ export const supportsColor: ColorInfo;
...
+/**
+@deprecated Use `BackgroundColorName` instead.
+
+Basic background color names.
+
+[More colors here.](https://github.com/chalk/chalk/blob/main/readme.md#256-and-truecolor-color-support)
+*/
+export type BackgroundColor = BackgroundColorName;
+
+/**
```
sindresorhus, line 304: Add TODO comment to remove this in the next major version.

sindresorhus (approved):

PR #572, opened 2022-10-18
Added support for CSS colors names

zenon8adams:
The feature adds the ability to translate CSS colour names into hex decodable format.


Qix-:
No thank, we used to have this with our use of `color-convert` and nobody used it. Just use `chalk.ansi256(colorConvert.keyword.ansi256('lavender'))` if you want to have this.

PR #581, opened 2022-12-08
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,10 @@ name: CI
 on:
   - push
   - pull_request
+
+permissions:
+  contents: read #  to fetch code (actions/checkout)
+
 jobs:
   test:
     name: Node.js ${{ matrix.node-version }}


sindresorhus:
Thanks for the PR, but this should be fixed by GitHub (they should fix their defaults). It's hopeless trying to add this to all repos. I'm also not interested in reviewing PRs for every repo.

sashashura:
Please note that you can [change the defaults per org or repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions).
image
It may break the existing workflows if they need write permissions and do not specify permissions explictly using the write-all defaults, but in your case it looks like all your workflows need read only.

Qix-:
I agree with Sindre's assertion here. Read-only should be the default - if it's a breaking change, so be it. This is Github's responsibility (and they've been lazy about proper security engineering for nearly a decade).

PR #600, opened 2023-06-14
Export `createSupportsColor` function from `supports-color`

lukekarrys:
I have a use for this function when using `chalk`. It would be nice if I could grab this function without needing to install `supports-color`, especially since it is vendored by `chalk`.

Edit: this also updates `xo` to the latest version to fix a crash when linting. Tests should hopefully pass on this PR now.


lmf-git:
@lukekarrys what is your use/what would be gained and what's the drawback? I'd guess 100 bytes...?

lukekarrys:
I hadn't even thought of the bytes. I see pros/cons for merging this as:

Cons:
- Maintenance burden, including needing to figure out semver for `chalk` if `supports-color` has breaking changes

Pros:
- Users only need to install and keep up to date 1 package
- Guarantee that version of `supports-color` is the same between `chalk` and one consumed by user

This is a very trivial change, and I don't mind if it's not merged. It would get used in my project, but I'm also fine using `supports-color` as well.

lmf-git:
Seems like a good contribution hope author replies and accepts it

mScott224 (approved):

PR #618, opened 2023-11-01
Remove Readme.md from bundle

### LINKED ISSUE #617: Remove README.md from bundle
The package is 44.7kb, and the README alone is 11.3kb.

~If you just remove the README from the bundle, you can save \~25% in bundle size, which could save 3TB (279,859,270 * 11.3kb) each week.~

We cannot remove but if we put a placeholder, we can save ~631GB/week (3,7×228240517÷1024÷1024) of bandwidth (13.3kB to 10.4kB).

H4ad:
Closes #617 

This will save ~631GB/week (3,7×228240517÷1024÷1024) of bandwidth (13.3kB to 10.4kB)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
 	"scripts": {
 		"test": "xo && c8 ava && tsd",
 		"bench": "matcha benchmark.js",
-		"prepack": "node -p 'require(`fs`).renameSync(`./readme.md`, `./_readme.md`)'",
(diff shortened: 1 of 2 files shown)


H4ad:
I think we should not remove it, but instead replace it with a cheaper version, removing it will probably make the page on NPM be empty.

sindresorhus:
https://github.com/chalk/chalk/issues/617

Issue #587, opened 2023-01-28
Replit demo not available


dzmitry-duboyski:
If I understand correctly, the link "Run on Replit" which is in README.md the file has an incorrect address.

In my opinion, this link should be in the following format: `https://replit.com/@user-name/project-name#folder/file-name.js`

When I click the current link https://repl.it/github/chalk/chalk , I don't see an example, but I see this page:
![chalk](https://user-images.githubusercontent.com/38065632/215270482-e9b7dbf2-bda7-443a-bdba-9f36894e832e.gif)

Correct me if I'm wrong. Thank you.

Qix-:
ReplIt has changed things I guess, the link used to work. It might also be requiring you to log in first.

so1ve:
You should login first

sygint:
Looks like this isn't the desired behavior at all

![Peek 2023-08-15 21-22](https://github.com/chalk/chalk/assets/872633/48b6f002-4e63-48f7-b71a-309447e02c6a)

Qix-:
WTH it's like they completely changed the entire site. This has no semblance to what it used to be...

@sindresorhus should we just remove it from the README? It seems they're having an [identity crisis](https://replit.com/talk/ask/what-happened-to-replit/121581) anyway..

so1ve:
Shall we run on stackblitz?

SamVerschueren:
> Shall we run on stackblitz?

Would be down for that 🙈

PR #622, opened 2024-01-06
Remove `.replit` file

hyperupcall:
As noted in #587, the Replit demo no longer works anymore. f7b29ae8ef4fd2048e08aa361778d290ed10ce7a partially fixes this, but the `.replit` file still remains. Even importing the GitHub project manually results in an error:

image

When the original PR was made, iirc dependencies were installed automatically. There are other things to note, too. In the `.replit` file, there is a similar `deployments.run` property which has similar functionality. As mentioned in the original issue, using replit to run the example now requires an account. I think these issues, along with the removal of the badge, suggests that the `.replit` file should be removed. Besides, there are better alternatives, like RunKit, or StackBlitz as mentioned previously.

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,2 +1,0 @@
-language = "nodejs"
-run = "cd examples && node rainbow"


Qix-:
Thanks, yeah I think this was missed.

Cheers!

PR #637, opened 2024-12-10
Update dependents count

akshar-dave:


sindresorhus:
Thanks, but we don't plan to constantly update this.

PR #640, opened 2024-12-20
Update `supports-color` to v10.0.0

### LINKED ISSUE #639: Not handling navigator being undefined
When updating to a 5.x-package, there is a problem when building our app.  It seems that the  current `supports-color` package is not able to handle `navigator` being `undefined`. This issue is fixed in the latest `supports-color` version `10.0.0`, so this can hopefully be fixed by updating this package :)

hyperupcall:
As mentioned in #639, the vendored `supports-color` only properly handles `navigator` being `undefined` since [this commit](https://github.com/chalk/supports-color/commit/c214314a14bcb174b12b3014b2b0a8de375029ae), which was released in version [v10.0.0](https://github.com/chalk/supports-color/releases/tag/v10.0.0).

This updates the vendored `supports-color` to `v10.0.0` to fix that issue.

Initially, I copied the files over byte-for-byte and tried to fix errors with dependencies and CI. That was very silly, so I've simply applied the required changes from `supports-color` (in a different branch).

Closes #639


package.json
```diff
@@ -18,7 +18,7 @@
 	"types": "./source/index.d.ts",
 	"sideEffects": false,
 	"engines": {
-		"node": "^12.17.0 || ^14.13 || >=16.0.0"
+		"node": ">=18"
```
sindresorhus, line 21: We don't plan to do a major release for Node.js 18 support. It's not worth having to do a major release.

hyperupcall: Oops, you're totally right! I have simplified my approach in #642, which supersedes this PR

hyperupcall:
This is superseded by #642

PR #657, opened 2025-09-09
check version and use built in replace all

ororsatti:
Based on the TODO on the code.
Check Node version and if 16 use built-in `replaceAll`.

Tested on
- Node 12
- Node 14
- Node > 16
-  Bun.


sindresorhus:
No thanks :)

PR #672, opened 2026-02-13
Fix FORCE_COLOR to set exact level instead of minimum

### LINKED ISSUE #624: $FORCE_COLOR works only as either level 0 or 3

veeceey:
When `FORCE_COLOR` is set to a specific level like `1` or `2`, the expected behavior is that the color support level is forced to exactly that value. However, `FORCE_COLOR` was being used as a floor/minimum value in `_supportsColor`, meaning terminal capability detection could still return a higher level.

For example, on a terminal with `COLORTERM=truecolor`, setting `FORCE_COLOR=1` would still give level 3 because the truecolor check fires after the minimum is set.

The fix makes `_supportsColor` return the forced level directly when `FORCE_COLOR` is explicitly set, skipping the terminal detection logic entirely. This matches the expected semantics — "force" means force to this exact level, not "ensure at least this level."

**Before:**
```
FORCE_COLOR=1 (with COLORTERM=truecolor) => level 3
FORCE_COLOR=2 (with COLORTERM=truecolor) => level 3
```

**After:**
```
FORCE_COLOR=1 (with COLORTERM=truecolor) => level 1
FORCE_COLOR=2 (with COLORTERM=truecolor) => level 2
```

Fixes #624

### DIFF
--- a/source/vendor/supports-color/index.js
+++ b/source/vendor/supports-color/index.js
@@ -91,7 +91,15 @@ function _supportsColor(haveStream, {streamIsTTY, sniffFlags = true} = {}) {
 		return 0;
 	}
 
-	const min = forceColor || 0;
+	// When `FORCE_COLOR` is set to a specific level, return that level
+	// directly instead of using it as a minimum. This ensures that
+	// e.g. FORCE_COLOR=1 gives exactly level 1, not a higher level
+	// based on terminal capabilities.
+	if (forceColor !== undefined) {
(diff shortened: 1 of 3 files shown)


sindresorhus:
https://github.com/sindresorhus/got/pull/2434#issuecomment-3895784376

PR #673, opened 2026-02-21
Fix security and performance issues (S2–S6, P1, P4, P6)

ayushkumar912:
## Summary

This PR addresses a set of security correctness issues and performance improvements found during a review of the codebase. All 35 existing tests continue to pass; three new regression tests are added.

### Security fixes

- **S2 — `hexToRgb` explicit type handling** (`ansi-styles/index.js`): The previous `hex.toString(16)` call was a no-op when `hex` was a string (strings ignore the radix argument), making the numeric-input path work only by accident. Now branches explicitly on `typeof hex === 'number'` and applies `.padStart(6, '0')` so leading zeros are preserved (e.g. `0x0000FF` → `"0000ff"` not `"ff"`).

- **S3 — `FORCE_COLOR` NaN propagation** (`supports-color/index.js`): `FORCE_COLOR=yes` (or any non-numeric string) caused `parseInt` to return `NaN`, which propagated through `Math.min` and `translateLevel` into a partially-truthy level object with `level: NaN`. Added `Number.isNaN` guard; invalid values now fall back to level `1`.

- **S4 — `_supportsColor` shared mutable state** (`supports-color/index.js`): `_supportsColor` had a hidden side effect — it overwrote the module-level `flagForceColor` variable (set from `process.argv` at load time) with the env-based value on every call. This could corrupt state across multiple calls in test environments that mock env vars. The effective value is now computed locally with no mutation.

- **S5 — `applyOptions` validation** (`index.js`): The guard `if (options.level && …)` short-circuited for any falsy value, so `null` and `NaN` bypassed validation and were stored as the level. Changed to `options.level !== undefined` so these values are properly rejected.

- **S6 — Browser Chromium version comparison** (`browser.js`): `brand.version > 93` relied on implicit string-to-number coercion. The UA spec returns `version` as a string; added explicit `Number(brand.version)` to make the intent clear and safe.

### Performance / correctness fixes

- **P1 — Stateful regex bug in `rainbow.js`** (`examples/rainbow.js`): The global-flagged regex `ignoreChars` was used with `.test()` inside a loop. Because `RegExp.prototype.test()` advances `lastIndex` after each match, consecutive non-printable characters (e.g. double-space, `\t\n`) were alternately misclassified — every other non-printable was incorrectly treated as printable. Replaced with a direct code-point comparison (`character < '!' || character > '~'`), which is also faster.

- **P4 — String concatenation in `stringEncaseCRLFWithFirstIndex`** (`utilities.js`): The loop used `returnValue +=` which allocates a new string on every iteration. Switched to collecting parts in an array and joining once at the end, reducing intermediate allocations for strings with many newlines.

- **P6 — Tagged template literals with interpolations** (`index.js`): `` chalk.red`hello ${name}` `` previously produced incorrect output (`"hello , World"`) because the builder treated the template strings array and interpolated values as flat `join(' ')` arguments. The builder now detects a tagged template call via the `.raw` property on the first argument and routes through `String.raw()` for correct interleaving.

## Test plan

- [x] All 35 pre-existing tests pass (`npm test`)
- [x] New test: `new Chalk({level: null})` and `new Chalk({level: NaN})` throw (S5)
- [x] New test: `hex(0xFF0000)`, `hex(0x0000FF)`, `hex(0x00FF00)` match their string equivalents including leading-zero case (S2)
- [x] New test: `` chalk.red`hello ${name}` `` and `` chalk.bold`count: ${42}` `` produce correctly styled output; plain array arguments unaffected (P6)
- [x] Lint clean (xo) — 0 errors, 6 pre-existing warnings (all in upstream TODO comments or pre-existing complexity)
- [x] Type checks pass (tsd)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/rainbow.js
+++ b/examples/rainbow.js
@@ -15,7 +15,7 @@ function rainbow(string, offset) {
 	let hue = offset % 360;
 	const characters = [];
 	for (const character of string) {
-		if (ignoreChars.test(character)) {
+		if (character < '!' || character > '~') {
 			characters.push(character);
 		} else {
 			characters.push(chalk.hex(convertColor.hsl.hex(hue, 100, 50))(character));
--- a/source/index.js
+++ b/source/index.js
@@ -22,7 +22,7 @@ const levelMapping = [
 const styles = Object.create(null);
 
 const applyOptions = (object, options = {}) => {
-	if (options.level && !(Number.isInteger(options.level) && options.level >= 0 && options.level <= 3)) {
+	if (options.level !== undefined && !(Number.isInteger(options.level) && options.level >= 0 && options.level <= 3)) {
 		throw new Error('The `level` option should be an integer from 0 to 3');
 	}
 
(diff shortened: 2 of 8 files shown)


sindresorhus:
No thanks. We don't accept pure AI PRs.

PR #687, opened 2026-07-23
Downsample `ansi256()` and `bgAnsi256()` to 16 colors at level 1

### LINKED ISSUE #686: ansi256()/bgAnsi256() skip ANSI-16 downsampling at level 1, unlike rgb()/hex() (patch and regression test ready)
Hi, thanks for maintaining chalk - it is one of the most widely used and best-kept small libraries in the ecosystem, and that reputation is exactly why we picked it for a full-rigor audit: not to hunt for problems, but to see whether a genuinely well-run project comes back clean. At HEAD `aa06bb5` (chalk 5.6.2, Node 22.18.0) the core library did come back clean on architecture, security, and correctness - `npm test` was green (32/32 ava tests, 99.61 percent statement coverage), xo and tsd both passed. One correctness bug turned up anyway, in the public API, and it seems worth your time: at color level 1, `ansi256()`/`bgAnsi256()` emit raw 256-color escape codes instead of downsampling to ANSI-16 the way `rgb()`/`hex()` already do. The suite did not catch it in part because that code path had zero test coverage before this audit (index.js lines 90-91 were never exercised); once coverage was added, a follow-up pass surfaced the downconversion gap itself. We have a small patch and a regression test ready, and we are glad to share it here or open a PR, whichever is easier on your end.

### `ansi256()`/`bgAnsi256()` don't downsample at level 1 (Medium, correctness)

Readme line 228 states: "Colors are downsampled from 16 million RGB values to an ANSI color format that is supported by the terminal emulator (or by specifying `{level: n}` as a Chalk option)." That holds for `rgb()` and `hex()`, but not for `ansi256()`/`bgAnsi256()`.

```
node -e "const {Chalk} = require('chalk'); const c = new Chalk({level: 1}); console.log(JSON.stringify(c.ansi256(196)('x')))"
```

At level 1 this prints `"\u001B[38;5;196mx\u001B[39m"` - the raw 256-color code, which a level-1 terminal does not support - while the equivalent call through the rgb model, `new Chalk({level: 1}).rgb(255, 0, 0)('x')`, correctly downsamples to `"\u001B[91mx\u001B[39m"`. `bgAnsi256(196)` shows the same gap against `bgRgb(255, 0, 0)` (raw `48;5;196` instead of the downsampled `101`). Level 2 and level 3 output is unaffected either way; this is specific to level-1 (basic, 16-color) terminals.

Root cause: in `source/index.js`, `getModelAnsi` branches the `rgb` model on level - `ansi16m` gets full truecolor, `ansi256` gets `rgbToAnsi256`, and anything else (level 1) gets downsampled further via `rgbToAnsi` - and the `hex` model redirects into that same `rgb` branch. The `ansi256` model has no such branch at all: every call falls through to the function's final, unconditional line, `ansiStyles[type][model](...arguments_)`, i.e. `ansiStyles[type].ansi256(...)`, regardless of level. `ansiStyles.ansi256ToAnsi` already exists in the vendored ansi-styles package and is exactly the conversion the `rgb` branch's own downsample step uses in spirit. The fix is one additional branch in `getModelAnsi`, checked before that fallback, mirroring the shape of the existing rgb-to-ansi branch: when `model === 'ansi256'` and `level === 'ansi'`, return `ansiStyles[type].ansi(ansiStyles.ansi256ToAnsi(...arguments_))` instead of falling through. With that branch in place, `ansi256(196)` at level 1 emits `"\u001B[91m...\u001B[39m"` and `bgAnsi256(196)` emits `"\u001B[101m...\u001B[49m"`, matching the rgb/hex behavior; level 2 and level 3 passthrough is unchanged.

The rest of what the audit found and fixed sits entirely in the tooling perimeter around the library, not the library itself: a dead `benchmark.js` that imported a file that no longer exists and depended on an abandoned `matcha` package, 12 to 14 known vulnerabilities in the xo/ava dev-dependency chains (zero in the shipped runtime, which has no runtime dependencies at all), the ansi256 test-coverage gap mentioned above, and a CI matrix that only tested EOL Node versions (14/16/18) with a deprecated `codecov-action@v2`. All of that is fixed in the same patch if you want it, but we want to be clear the core library came through the full audit clean, which given chalk's reputation was the result we expected going in.

### What we have ready

A patch against `aa06bb5` (`source/index.js`, `test/chalk.js`, plus the tooling-perimeter changes above - 6 files, +76/-67, excluding the regenerated `package-lock.json`) adds the one `getModelAnsi` branch and a regression test covering `ansi256` and `bgAnsi256` at level 1, and confirms level 2/3 output is unchanged. With the full patch applied: 34/34 tests passing, 100 percent statement coverage, `npm audit` at 0 vulnerabilities. We are happy to attach or link the patch, whichever is easier for you to review.

This finding comes out of a documented audit run - journal, patch, and report are published at https://github.com/lenamonj/jeffy-loop/tree/main/evals/chalk - and it was reproduced and reviewed by a human before we filed this.

If it would help, we are glad to open a PR for just the `ansi256`/`bgAnsi256` fix, or for the wider tooling patch, whichever you would prefer - and happy to follow any contribution conventions you would like us to use.

sindresorhus:
At color level 1, `ansi256()`/`bgAnsi256()` emitted raw 256-color escape codes that a basic 16-color terminal cannot render, while `rgb()`/`hex()` downsampled correctly.

Fixes #686

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -225,7 +225,7 @@ console.log(foregroundColorNames.includes('pink'));
 
 Chalk supports 256 colors and [Truecolor](https://github.com/termstandard/colors) (16 million colors) on supported terminal apps.
 
-Colors are downsampled from 16 million RGB values to an ANSI color format that is supported by the terminal emulator (or by specifying `{level: n}` as a Chalk option). For example, Chalk configured to run at level 1 (basic color support) will downsample an RGB value of #FF0000 (red) to 91 (ANSI escape for bright red).
+Colors are downsampled from 16 million RGB values to an ANSI color format that is supported by the terminal emulator (or by specifying `{level: n}` as a Chalk option). For example, Chalk configured to run at level 1 (basic color support) will downsample an RGB value of #FF0000 (red) to 91 (ANSI escape for bright red). The same applies to `ansi256` values, so `chalk.ansi256(196)` also becomes 91 at level 1.
 
 Examples:
 
--- a/source/index.d.ts
+++ b/source/index.d.ts
@@ -74,6 +74,8 @@ export interface ChalkInstance {
 	/**
 	Use an [8-bit unsigned number](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) to set text color.
 
+	The value is downsampled to the 16-color palette on terminals that only support basic colors (level 1), so `chalk.ansi256(196)` becomes 91 (ANSI escape for bright red).
+
 	@example
 	```
 	import chalk from 'chalk';
@@ -111,6 +113,8 @@ export interface ChalkInstance {
 
 	/**
 	Use an [8-bit unsigned number](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) to set background color.
+
+	The value is downsampled to the 16-color palette on terminals that only support basic colors (level 1), so `chalk.bgAnsi256(196)` becomes 101 (ANSI escape for bright red background).
 
 	@example
 	```
--- a/source/index.js
+++ b/source/index.js
@@ -88,6 +88,10 @@ const getModelAnsi = (model, level, type, ...arguments_) => {
 		return getModelAnsi('rgb', level, type, ...ansiStyles.hexToRgb(...arguments_));
 	}
 
+	if (model === 'ansi256' && level === 'ansi') {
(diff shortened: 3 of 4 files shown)


Qix- (approved):

Qix-:
LGTM

